PR #41, opened 2020-07-01
fix typo in polling source

vsinghal13:


website/docs/r/polling_source.html.markdown
```diff
@@ -76,7 +76,7 @@ resource "sumologic_polling_source" "terraform_cw_metrics" {
     limit_to_namespaces = ["AWS/Route53","AWS/S3","customNamespace"]
   
     dynamic "tag_filters" {
-    for_each = local.filters
+    for_each = local.tagfilters
     content{
```
mccartney, line 80:
```suggestion
    content {
```

mccartney:
Also an extra `}` in line 87 (cannot submit a suggesting comment in Github)

vsinghal13:
> Also an extra `}` in line 87 (cannot submit a suggesting comment in Github)

why do we need an extra `}` there?

mccartney:
> why do we need an extra } there?

I don't think we need it. It seems like we have 1 too many.

PR #45, opened 2020-07-08
use the official function for json diff

seansain:


seansain:
No, that seems to be a problem with the API

PR #47, opened 2020-07-10
content/folder: use resource.StateChangeConf to improve job polling

bendrucker:
This PR introduces  [`resource.StateChangeConf`](https://www.terraform.io/docs/extend/resources/retries-and-customizable-timeouts.html#statechangeconf) to help improve the reliability and messaging for library management resources with a job status API. 

It adds customizable timeouts for each resource/operation that has job polling and passes them to `StateChangeConf`, which provides:

* Handling of errors: previously, a "Failed" job resulted in an infinite loop. We observed this at @takescoop which is what brought me here. Now "Failed" is an unknown state, and will trigger an immediate error.
* Timeouts for jobs that don't progress: previously, an "InProgress" job that never became successful resulted in an infinite loop. We didn't observe this, but it's worth defending against anyway. 
* Backoff: previously, requests were issued every second. With backoff, far fewer requests will be sent for long-running jobs (hopefully reduces pressure on the rate limiter), with minimal delay for users.

`StateChangeConf` helps with readable logs/errors for debugging as well:

```
errors during refresh: timeout while waiting for state to become 'Success' (last state: 'InProgress', timeout: 1m0s)
```

## Acceptance Tests

```
$ TESTARGS="-run 'TestAcc(Content|Folder)'" make testacc
==> Checking that code complies with gofmt requirements...
TF_ACC=1 go test $(go list ./... |grep -v 'vendor') -v -run 'TestAcc(Content|Folder)' -timeout 120m
?   	github.com/terraform-providers/terraform-provider-sumologic	[no test files]
=== RUN   TestAccContent_create
--- PASS: TestAccContent_create (35.05s)
=== RUN   TestAccContent_update
--- PASS: TestAccContent_update (67.27s)
=== RUN   TestAccFolder_create
--- PASS: TestAccFolder_create (7.73s)
=== RUN   TestAccFolder_update
--- PASS: TestAccFolder_update (9.51s)
PASS
ok  	github.com/terraform-providers/terraform-provider-sumologic/sumologic
```

## Questions

* Do these timeouts make sense? I'm guessing, but inside knowledge of the API behavior (e.g. 99% latency for jobs) would help inform the value 🙂.

### DIFF
--- a/sumologic/resource_sumologic_content.go
+++ b/sumologic/resource_sumologic_content.go
@@ -31,9 +31,9 @@ func resourceSumologicContent() *schema.Resource {
 			},
 		},
 		Timeouts: &schema.ResourceTimeout{
-			Read:   schema.DefaultTimeout(time.Duration(1) * time.Minute),
-			Create: schema.DefaultTimeout(time.Duration(10) * time.Minute),
-			Delete: schema.DefaultTimeout(time.Duration(1) * time.Minute),
+			Read:   schema.DefaultTimeout(1 * time.Minute),
+			Create: schema.DefaultTimeout(10 * time.Minute),
+			Delete: schema.DefaultTimeout(1 * time.Minute),
 		},
 	}
 }
--- a/sumologic/resource_sumologic_content.go
+++ b/sumologic/resource_sumologic_content.go
@@ -3,6 +3,7 @@ package sumologic
 import (
 	"encoding/json"
 	"log"
+	"time"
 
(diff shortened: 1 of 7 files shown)


bendrucker:
If there's a consistent way to trigger errors, I'd also be happy to add an acceptance test against the "Failed" job handling.

seansain:
@bendrucker Thanks for this PR! This should definitely improve our async APIs!

PR #58, opened 2020-07-24
update content resource to allow updates using overwrite flag

seansain:
- recreating content items from the Terraform Provider deletes links to dashboards
- adding update using the overwrite API flag
- this creates a bug when changing the Name - DO NOT change the name of a resource, just delete and recreate if you have to change the name

### DIFF
--- a/sumologic/sumologic_client.go
+++ b/sumologic/sumologic_client.go
@@ -42,7 +42,7 @@ func createNewRequest(method, url string, body io.Reader, accessID string, acces
 		return nil, err
 	}
 	req.Header.Add("Content-Type", "application/json")
-	req.Header.Add("User-Agent", "SumoLogicTerraformProvider/2.0.2")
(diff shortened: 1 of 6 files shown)


sumologic/sumologic_client.go
```diff
@@ -42,7 +42,7 @@ func createNewRequest(method, url string, body io.Reader, accessID string, acces
 		return nil, err
 	}
 	req.Header.Add("Content-Type", "application/json")
-	req.Header.Add("User-Agent", "SumoLogicTerraformProvider/2.0.2")
+	req.Header.Add("User-Agent", "SumoLogicTerraformProvider/2.1.3")
```
sumovishal, line 45: We should find a better way to do this. This should be updated for each release.

seansain: yup, I've pretty much forgotten every release.

Issue #46, opened 2020-07-10
Feature request: AWS Metadata (Tag) Source support


mccartney:
I'd be great if we could use Terraform to define this kind of source: https://help.sumologic.com/03Send-Data/Sources/02Sources-for-Hosted-Collectors/Amazon-Web-Services/AWS-Metadata-(Tag)-Source

PR #61, opened 2020-08-07
Add support for AWS Metadata Source

vsinghal13:
Solves #46 

Eg config:
```
resource "sumologic_metadata_source" "tf_metadata" {
  name          = "Amazon Metadata1"
  category      = "aws/metadata"
  content_type  = "AwsMetadata"
  scan_interval = 300000
  paused        = false

  authentication {
    type = "AWSRoleBasedAuthentication"
    role_arn = "xxx"
  }

  path {
    type = "AwsMetadataPath"
    limit_to_regions = ["us-west-2"]
    limit_to_namespaces = ["AWS/EC2"]
    tag_filters = ["Deploy*","Cluster", "!DeployStatus"]
  }
}

```
Have created a new resource because the `tag_filters` for this source type accepts a list of strings as compared to `tag_filters` of polling source which is a list of maps with `type`, `namespace` and `tags` as keys.

Also tested CRUD functionality of this resource locally.

### DIFF
--- a/sumologic/resource_sumologic_metadata_source.go
+++ b/sumologic/resource_sumologic_metadata_source.go
@@ -75,15 +75,7 @@ func resourceSumologicMetadataSource() *schema.Resource {
 				"type": {
 					Type:         schema.TypeString,
 					Required:     true,
-					ValidateFunc: validation.StringInSlice([]string{"S3BucketPathExpression", "CloudWatchPath", "AwsMetadataPath"}, false),
-				},
-				"bucket_name": {
-					Type:     schema.TypeString,
-					Optional: true,
-				},
-				"path_expression": {
-					Type:     schema.TypeString,
-					Optional: true,
+					ValidateFunc: validation.StringInSlice([]string{"AwsMetadataPath"}, false),
 				},
 				"limit_to_regions": {
 					Type:     schema.TypeList,
@@ -214,8 +206,6 @@ func getMetadataThirdPartyPathAttributes(pollingResource []MetadataResource) []m
 	for _, t := range pollingResource {
 		mapping := map[string]interface{}{
 			"type":                t.Path.Type,
-			"bucket_name":         t.Path.BucketName,
-			"path_expression":     t.Path.PathExpression,
 			"limit_to_regions":    t.Path.LimitToRegions,
 			"limit_to_namespaces": t.Path.LimitToNamespaces,
 			"tag_filters":         t.Path.TagFilters,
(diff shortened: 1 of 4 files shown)


website/docs/r/metadata_source.html.markdown
```diff
@@ -0,0 +1,86 @@
+---
+layout: "sumologic"
+page_title: "SumoLogic: sumologic_metadata_source"
+description: |-
+  Provides a Sumologic Metadata (Tag) source. This source allows you to collect tags from EC2 instances running on AWS.
+---
+
+# sumologic_polling_source
```
frankreno, line 8: Should this be metadata?

frankreno: The header

maimaisie: not sure if this was discussed earlier, did we decide on adding the metadata source as a separate source resource instead of to the polling source like what we did for cloudwatch?

vsinghal13: Yes, as per the earlier discussion it was decided to add separate resources for all AWS sources so that we have 1:1 mapping with what we show on UI. 

Also, as mentioned in the description
>Have created a new resource because the tag_filters for this source type accepts a list of strings as compared to tag_filters of polling source which is a list of maps with type, namespace and tags as keys.

vsinghal13: I am working on splitting the `polling_source` into separate sources and will create a new PR for that.

maimaisie: thanks for the clarification

sumologic/resource_sumologic_metadata_source.go
```diff
@@ -0,0 +1,279 @@
...
+		MinItems: 1,
+		MaxItems: 1,
+		Elem: &schema.Resource{
+			Schema: map[string]*schema.Schema{
+				"type": {
+					Type:         schema.TypeString,
+					Required:     true,
+					ValidateFunc: validation.StringInSlice([]string{"S3BucketPathExpression", "CloudWatchPath", "AwsMetadataPath"}, false),
+				},
+				"bucket_name": {
```
maimaisie, line 80: bucket name and path expression are not needed by the AWS metadata source, why do we add them in the schema?

vsinghal13: good catch! will fix that.

seansain:
We really need to get some acceptance tests in for these. You can just create a test bucket and store the role ARN as an environment variable in Travis. We used to do a similar thing with the personal folder ID.

PR #69, opened 2020-08-12
terraform support for inventory source

apoorv-garg:


apoorv-garg:
Please feel free to review, while I add acceptance tests

apoorv-garg:
@seansain to write acceptance tests, we would need to hardcode an AWS role ARN, or create one dynamically using cloudformation, in which case we would need to hardcode some aws credentials. Is there a prior art for this ? I can't find any existing tests for the polling source.

Also please let me know if acceptance tests are an absolute must here. We are running on a very tight GA timeline, so another option would be to add tests post that (1st week of september), and work with QE to release these without acceptance tests meanwhile.

seansain:
Ohh, that's right. I forgot polling source doesn't have acceptance tests yet. We can hard code a role ARN if we use it as environment variable. That would be ideal I think.

However, QE verification for the tight GA timeline should suffice for now.

seansain:
LGTM, but I think @maimaisie or @vsinghal13 could provide a better review.

seansain:
just merged #70 that should fix the lint error

website/docs/r/aws_inventory_source.html.markdown
```diff
@@ -0,0 +1,91 @@
...
+
+## Argument reference
+
+In addition to the common properties, the following arguments are supported:
+
+ - `content_type` - (Required) The content-type of the collected data. This has to be `AwsInventoryPath` for AWS Inventory source.
+ - `scan_interval` - (Required) Time interval in milliseconds of scans for new data. The minimum value is 1000 milliseconds. Currently this value is not respected.
+ - `paused` - (Required) When set to true, the scanner is paused. To disable, set to false.
+ - `authentication` - (Required) Authentication details to access AWS `Describe*` APIs.
+     + `type` - (Required) Must be either `S3BucketAuthentication` or `AWSRoleBasedAuthentication`
```
vsinghal13, line 50: Maybe we should modify this to have only `AWSRoleBasedAuthentication` as `S3BucketAuthentication` is not supported for AWS Inventory source?

sumologic/resource_sumologic_generic_polling_source.go
```diff
@@ -278,29 +293,35 @@ func getPollingTagFilters(d *schema.ResourceData) []TagFilter {
...
 	auths := d.Get("authentication").([]interface{})
 	authSettings := PollingAuthentication{}
 
 	if len(auths) > 0 {
 		auth := auths[0].(map[string]interface{})
 		switch authType := auth["type"].(string); authType {
 		case "S3BucketAuthentication":
+			if d.Get("content_type").(string) == "AwsInventory" {
+				return authSettings, errors.New(
+					fmt.Sprintf("[ERROR] Unsupported authType: %v for AwsInventory source", authType))
```
maimaisie, line 306: do we block source creation on the backend when the auth type is bucket auth for inventory sources? if so, I'm wondering if we need to repeat the validation in the tf layer

website/docs/r/aws_inventory_source.html.markdown
apoorv-garg: Good catch, fixed.

sumologic/resource_sumologic_generic_polling_source.go
apoorv-garg: We don't actually. We are, however, considering adding support in the backend, so if and when that is done, I will remove this check from here. OTOH, if we end up deciding not to support bucket auth in backend, we will make that check in the backend in addition to this one.

seansain:
@apoorv-garg can you update the Changelog with info on the two new PRs before merging?

apoorv-garg:
> @apoorv-garg can you update the Changelog with info on the two new PRs before merging?

Done

vsinghal13:
@apoorv-garg could you please add acceptance tests for Xray and Inventory sources? Please refer to #72 for templates.

apoorv-garg:
Sure, will try to do it by EOW

PR #72, opened 2020-08-20
add acceptance test for polling source

vsinghal13:


sumologic/resource_sumologic_cloudfront_source_test.go
```diff
@@ -0,0 +1,194 @@
...
+)
+
+func TestAccSumologicCloudFrontSource_create(t *testing.T) {
+	var cloudFrontSource PollingSource
+	var collector Collector
+	cName := acctest.RandomWithPrefix("tf-acc-test")
+	cDescription := acctest.RandomWithPrefix("tf-acc-test")
+	cCategory := acctest.RandomWithPrefix("tf-acc-test")
+	sName := acctest.RandomWithPrefix("tf-acc-test")
+	sDescription := acctest.RandomWithPrefix("tf-acc-test")
```
maimaisie, line 21: couple questions,
1. what's prefix `c` and `s`?
2. do we need randomized description and category?
3. these are all repeated in every test, can we DRY it?

vsinghal13: 1. `c` - collector, `s` - source
2 & 3: I followed the same pattern as present in other source acceptance tests.

PR #79, opened 2020-09-11
SUMO-147334: fix monitor state refresh

rsanbhad:
Fixes state refresh for the sub-types in monitors.


sumovishal:
tests are failing

GNUmakefile
```diff
@@ -68,4 +68,4 @@ ifeq (,$(wildcard $(GOPATH)/src/$(WEBSITE_REPO)))
 endif
 	@$(MAKE) -C $(GOPATH)/src/$(WEBSITE_REPO) website-provider-test PROVIDER_PATH=$(shell pwd) PROVIDER_NAME=$(PKG_NAME)
 
-.PHONY: build install uninstall test testacc vet fmt fmtcheck errcheck lint tools test-compile website website-lint website-test
\ No newline at end of file
+.PHONY: build install uninstall test testacc vet fmt fmtcheck errcheck lint tools test-compile website website-lint website-test
```
seansain, line 71: still showing a change in makefile

sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -291,6 +297,72 @@ func resourceSumologicMonitorsLibraryMonitorRead(d *schema.ResourceData, meta in
 	d.Set("is_disabled", monitor.IsDisabled)
 	d.Set("status", monitor.Status)
 	d.Set("group_notifications", monitor.GroupNotifications)
+	log.Printf("[DEBUG] monitor notifications: %v", monitor.Notifications)
+	log.Printf("[DEBUG] monitor triggers: %v", monitor.Triggers)
+	log.Printf("[DEBUG] monitor queries: %v", monitor.Queries)
```
seansain, line 302: do we still want these?

rsanbhad: removed

GNUmakefile
rsanbhad: reverted this change, but I notice this happens with all my PRs because vscode adds a newline whenever i open/save this file - it should be ok to add a newline to the makefile right?

seansain: huh, I use vscode too, but I've never had this problem. ya should be fine otherwise.

seansain:
Let's merge this? The acc tests seem to be broken on master rn.

seansain (approved):

PR #82, opened 2020-09-17
Add provider for custom fields API

mfatyga-sumo:


sumologic/resource_sumologic_field.go
```diff
@@ -0,0 +1,131 @@
...
+	"fmt"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceSumologicField() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceSumologicFieldCreate,
+		Read:   resourceSumologicFieldRead,
+		Update: resourceSumologicFieldUpdate,
```
seansain, line 13: Since update is `nil` we can simply remote it instead. Similar to how it was in content: https://github.com/SumoLogic/terraform-provider-sumologic/blob/5501c3e68ff12e2bdf6a03a7d227e7247d09fbe8/sumologic/resource_sumologic_content.go

seansain:
You will need to add this resource to the `provider.go` file for it to be picked up by the provider.

seansain:
Would be nice to get some docs too.

mfatyga-sumo:
> Would be nice to get some docs too.

What form of documentation do you think would be appropriate, godoc documentation or something else?

sumologic/resource_sumologic_field.go
```diff
@@ -0,0 +1,126 @@
...
+
+			"field_id": {
+				Type:     schema.TypeString,
+				Computed: true,
+				ForceNew: false,
+			},
+
+			"data_type": {
+				Type:     schema.TypeString,
+				Required: true,
```
frankreno, line 34: Why is this required? We don't require this on the create API call.

sumologic/resource_sumologic_field.go
```diff
@@ -0,0 +1,126 @@
...
+				ForceNew: true,
+			},
+
+			"state": {
+				Type:     schema.TypeString,
+				Required: true,
```
frankreno, line 40: Same comment as above...

frankreno (approved):
LGTM, pending docs and questions on the 2 required fields.

frankreno:
> > Would be nice to get some docs too.
> 
> What form of documentation do you think would be appropriate, godoc documentation or something else?

See the website/docs folder in root of readme

mfatyga-sumo:
Please take another look.

PR #87, opened 2020-09-20
Add Lookup Table resource

sumovishal:
Docs will be added in a separate PR.


sumologic/resource_sumologic_lookup_table.go
```diff
@@ -119,7 +122,7 @@ func resourceSumologicLookupTableRead(d *schema.ResourceData, meta interface{})
 
 	id := d.Id()
 	lookupTable, err := c.GetLookupTable(id)
-	fmt.Printf("##DEBUG## read lookup: %+v\n", lookupTable)
+	log.Printf("##DEBUG## read lookup: %+v\n", lookupTable)
```
sumovishal, line 125: You can remove these and other print statements I added. I added them for debugging purpose.

sumologic/resource_sumologic_lookup_table.go
```diff
@@ -131,7 +134,9 @@ func resourceSumologicLookupTableRead(d *schema.ResourceData, meta interface{})
 	}
 
 	d.Set("name", lookupTable.Name)
-	d.Set("fields", lookupTable.Fields)
+	if err := d.Set("fields", fieldsToList(lookupTable.Fields)); err != nil {
```
sumovishal, line 137: We need to fix codegen as it doesn't handle this correctly.

sumovishal (commented):
lgtm, you can remove the print statements.

sumologic/resource_sumologic_lookup_table.go
seansain: they only come if someone sets `TF_LOG` to `debug`. We can keep them for further debugging.

PR #94, opened 2020-10-13
fixing action_type connection_type compatibility

rsanbhad:
A backwards-compatible fix for an API change that we made recently in monitors, which will roll out to prods this week.

### DIFF
--- a/sumologic/resource_sumologic_monitors_library_monitor.go
+++ b/sumologic/resource_sumologic_monitors_library_monitor.go
@@ -308,7 +308,9 @@ func resourceSumologicMonitorsLibraryMonitorRead(d *schema.ResourceData, meta in
 		if internalNotificationDict["connectionType"] != nil {
 			internalNotification["connection_type"] = internalNotificationDict["connectionType"].(string)
 		}
-		internalNotification["action_type"] = internalNotificationDict["actionType"].(string)
+		if internalNotificationDict["actionType"] != nil {
+			internalNotification["action_type"] = internalNotificationDict["actionType"].(string)
+		}
 		if internalNotification["action_type"].(string) == "EmailAction" ||
 			internalNotification["action_type"].(string) == "Email" ||
 			internalNotification["connection_type"].(string) == "EmailAction" ||
@@ -398,7 +400,8 @@ func getNotifications(d *schema.ResourceData) []MonitorNotification {
 			notificationActionDict["connection_type"].(string) == "EmailAction" ||
 			notificationActionDict["connection_type"].(string) == "Email" {
 			notificationAction := EmailNotification{}
-			notificationAction.ActionType = notificationActionDict["action_type"].(string)
+			// notificationAction.ActionType = notificationActionDict["action_type"].(string)
+			notificationAction.ConnectionType = notificationActionDict["connection_type"].(string)
 			notificationAction.Subject = notificationActionDict["subject"].(string)
 			notificationAction.Recipients = notificationActionDict["recipients"].([]interface{})
 			notificationAction.MessageBody = notificationActionDict["message_body"].(string)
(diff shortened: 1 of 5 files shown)


rsanbhad:
acceptance tests seem to be failing on something unrelated 🤔 
```
=== RUN   TestAccContent_create
--- FAIL: TestAccContent_create (38.29s)
    testing.go:654: Step 0 error: errors during apply:
        
        Error: {"id":"RD347-56ON1-UIXWE","errors":[{"code":"request.url.invalid","message":"The request was rejected because URL is invalid."}]}
        
          on /tmp/tf-test019199049/main.tf line 3:
          (source code not available)
```
cc @seansain @sumovishal do you know who owns that test/what might have broken that?

sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -305,19 +307,29 @@ func resourceSumologicMonitorsLibraryMonitorRead(d *schema.ResourceData, meta in
...
 			internalNotification["connection_type"] = internalNotificationDict["connectionType"].(string)
+		} else {
+			// for backwards compatibility
+			internalNotification["connection_type"] = internalNotificationDict["actionType"].(string)
+			// convert from old action_type name to new connection_type name if applicable
+			if internalNotification["connection_type"].(string) == "EmailAction" {
+				internalNotification["connection_type"] = "Email"
+			}
+			if internalNotification["connection_type"].(string) == "NamedConnectionAction" {
+				internalNotification["connection_type"] = "Webhook"
```
tarunk2, line 321: This is not correct. `NamedConnectionAction` can be of type `Slack`, `Jira`, `PagerDuty`, `Webhook` etc.

sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -393,20 +405,39 @@ func getNotifications(d *schema.ResourceData) []MonitorNotification {
...
+		} else {
+			// for backwards compatibility
+			actionType = notificationActionDict["action_type"].(string)
+			connectionType = actionType
+			// convert from old action_type name to new connection_type name if applicable
+			if connectionType == "EmailAction" {
+				connectionType = "Email"
+			}
+			if connectionType == "NamedConnectionAction" {
+				connectionType = "Webhook"
```
tarunk2, line 423: Same here.

sumologic/resource_sumologic_monitors_library_monitor.go
rsanbhad: This is specifically for the case where we have existing `action_type = "NamedConnectionAction"` - in that case we don't have enough information from the resource itself to decide which of the webhook types it should be so only the generic "Webhook" connection type makes sense right? if the user has `action_type = "PagerDuty"` then connection_type will be "PagerDuty"

sumologic/resource_sumologic_monitors_library_monitor.go
rsanbhad: responded in above comment thread

sumologic/resource_sumologic_monitors_library_monitor.go
tarunk2: `NamedConnectionAction` allows all types right now. The good part is that we have done one round of migration and the next one is planned for soon so we are fine at DB layer. But if a customer has set up a monitor with this type for `slack`/`pagerduty` etc via Terraform, it will save it as `Webhook` on update after this change. There is no clean solution to make this truly backward compatible. We will have to work with customers to use `Slack` etc instead of `NamedConnectionAction`.

rsanbhad: Yup - also from what I tested, there doesn't seem to be an immediate functional impact to using "Webhook" for connection_type instead of the more specific types "PagerDuty"/"Slack" - the notification/payload still seems to shows up correctly in the UI and API

sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -135,10 +135,12 @@ func resourceSumologicMonitorsLibraryMonitor() *schema.Resource {
 									"action_type": {
 										Type:     schema.TypeString,
 										Optional: true,
+										Computed: true,
```
seansain, line 138: can we add a deprecation note to this attribute: https://www.terraform.io/docs/extend/best-practices/deprecations.html

seansain (approved):
lgtm otherwise - but add the deprecation note, to help customers transition to the new attribute

PR #92, opened 2020-10-12
Check for errors when getting collector from client

pmalek-sumo:
If the is not performed then on some occasions `collector` might be set to `nil` and then 

```go
d.SetId(strconv.Itoa(collector.ID))
```

will panic as below:

```
2020-10-12T09:25:33.603Z [DEBUG] plugin: plugin process exited: path=/terraform/.terraform/plugins/linux_amd64/terraform-provider-kubernetes_v1.11.4_x4 pid=110
2020-10-12T09:25:33.603Z [DEBUG] plugin: plugin exited
2020/10/12 09:25:33 [TRACE] [walkImport] Exiting eval tree: provider.kubernetes (close)
2020/10/12 09:25:33 [TRACE] vertex "provider.kubernetes (close)": visit complete
2020/10/12 09:25:33 [TRACE] <root>: eval: *terraform.EvalRefresh
2020/10/12 09:25:33 [TRACE] GRPCProvider: ReadResource
2020-10-12T09:25:33.750Z [DEBUG] plugin.terraform-provider-sumologic_v2.0.3_x4: panic: runtime error: invalid memory address or nil pointer dereference
2020-10-12T09:25:33.750Z [DEBUG] plugin.terraform-provider-sumologic_v2.0.3_x4: [signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xda406d]
2020-10-12T09:25:33.750Z [DEBUG] plugin.terraform-provider-sumologic_v2.0.3_x4:
2020-10-12T09:25:33.750Z [DEBUG] plugin.terraform-provider-sumologic_v2.0.3_x4: goroutine 46 [running]:
2020-10-12T09:25:33.750Z [DEBUG] plugin.terraform-provider-sumologic_v2.0.3_x4: github.com/terraform-providers/terraform-provider-sumologic/sumologic.resourceSumologicCollectorRead(0xc0001ce5b0, 0x102cb20, 0xc0004f60a0, 0xc0001ce5b0, 0x0)
2020-10-12T09:25:33.750Z [DEBUG] plugin.terraform-provider-sumologic_v2.0.3_x4:         /opt/teamcity-agent/work/5d79fe75d4460a2f/src/github.com/terraform-providers/terraform-provider-sumologic/sumologic/resource_sumologic_collector.go:77 +0x11d
[...]

PR #99, opened 2020-10-16
Log when received collector is nil

pmalek-sumo:


sumologic/resource_sumologic_collector.go
```diff
@@ -61,16 +61,15 @@ func resourceSumologicCollectorRead(d *schema.ResourceData, meta interface{}) er
 		collector, err = c.GetCollectorName(d.Id())
 		if err != nil {
 			log.Printf("[WARN] Collector not found when looking by name: %s, err: %v", d.Id(), err)
-			d.SetId("")
-			return err
+		} else if collector == nil {
+			log.Printf("[WARN] Got a nil Collector when looking by name: %s", d.Id())
```
rsanbhad, line 65: I think we still need `d.SetId("")` here? otherwise the provider will not clear state on refresh for resources that no longer exist

pmalek-sumo: I might not fully understand the domain here but what I'm trying to do here is to revert to version from before my last "fix" (https://github.com/SumoLogic/terraform-provider-sumologic/pull/92) - snippet below:

```go
        id, err := strconv.Atoi(d.Id())

        var collector *Collector
        if err != nil {
                collector, _ = c.GetCollectorName(d.Id())
                d.SetId(strconv.Itoa(collector.ID))
        } else {
                collector, _ = c.GetCollector(id)
        }

        if collector == nil {
                log.Printf("[WARN] Collector not found, removing from state: %v - %v", id, err)
                d.SetId("")

                return nil
        }
```

and:
* add some logging
* prevent from `panic`ing when `collector == nil` and we tried to call `d.SetId(strconv.Itoa(collector.ID))`, the `collector.ID` part is the one that would `panic`

The diff might not be 100% accurate here to reflect what was here before.

pmalek-sumo: Just to call this out explicitly: with the version from this PR we don't return in those `if`s hence we'll fall through to lines below which will do that:

```go
	if collector == nil {
		log.Printf("[WARN] Collector not found, removing from state: %v - %v", id, err)
		d.SetId("")
		return nil
	}
```

rsanbhad: ah ok - my bad, ya seems I was reading the diff wrong 🤔

vsinghal13 (approved):

maimaisie (approved):

PR #119, opened 2020-12-09
Fix analytics_tier values for partition resource

sumovishal:
Also fixed file permissions.


sumologic/resource_sumologic_partition.go
```diff
@@ -29,7 +29,7 @@ func resourceSumologicPartition() *schema.Resource {
 			"analytics_tier": {
 				Type:         schema.TypeString,
 				Optional:     true,
-				ValidateFunc: validation.StringInSlice([]string{"enhanced", "basic", "cold"}, false),
+				ValidateFunc: validation.StringInSlice([]string{"continuous", "frequent", "infrequent"}, false),
 				Default:      "enhanced",
```
seansain, line 33: Default shouldn't be enhanced then right? 
And I remember seeing somewhere that the API is backwards compatible so Terraform should accept old and new probably.

msuterski:
Hey. I run into this problem with missing tier types myself today. Any update on this PR?

seansain:
there are empty files in here going to close and replace with #148

PR #120, opened 2020-12-10
Add sumologic gcp source

vsinghal13:
Adds a new GCP source.
- does not explicitly require `content_type` as it is set to `GoogleCloudLogs` by default.
- does not need `authentication` and `path` blocks.

Sample Config: 
```
resource "sumologic_gcp_source" "terraform_gcp_source" {
  name          = "GCP Source"
  category      = "gcp"
}

resource "sumologic_collector" "collector" {
  name        = "my-collector"
  description = "Just testing this"
}
```


website/docs/r/gcp_source.html.markdown
```diff
@@ -0,0 +1,50 @@
+---
+layout: "sumologic"
+page_title: "SumoLogic: sumologic_gcp_source"
+description: |-
+  Provides a Sumologic GCP Source.
+---
+
+# sumologic_gcp_source
+Provides a [Sumologic GCP Source][2].
```
maimaisie, line 9: Can we use the full name (not abbreviation) here? (should at least use the full name once on this page)

maimaisie: Also, `Sumo Logic` not `Sumologic` probably other pages have this wrong as well

maimaisie (approved):

PR #129, opened 2020-12-21
Add test case for Kinesis log source

Gourav2906:
Adding test case for kinesis log similar to tracing to improve test coverage


vsinghal13 (approved):
Do we also want to update the documentation with the new `content_type` for `http` source?

maimaisie (approved):

PR #131, opened 2020-12-23
Add isActive field to handle decommissioned partitions

sumovishal:
The new way to look for decommissioned partitions is via the `isActive` field. Adding it to the resource to handle decommissioned partitions.


seansain (approved):
Lgtm

PR #135, opened 2021-01-15
Add authentication and path params in gcp example

vsinghal13:
Looks like when the optional fields are not specified, they are set to `null` even though we have a default value for them in code. So modifying the example to explicitly set the fields.

### DIFF
--- a/website/docs/r/gcp_source.html.markdown
+++ b/website/docs/r/gcp_source.html.markdown
@@ -16,6 +16,13 @@ resource "sumologic_gcp_source" "terraform_gcp_source" {
   description   = "My description"
   category      = "gcp"
   collector_id  = "${sumologic_collector.collector.id}"
+  authentication {
+    type = "NoAuthentication"
+  }
+
+  path {
+    type = "NoPathExpression"
+  }
 }
 
 resource "sumologic_collector" "collector" {


website/docs/r/gcp_source.html.markdown
```diff
@@ -16,6 +16,13 @@ resource "sumologic_gcp_source" "terraform_gcp_source" {
   description   = "My description"
   category      = "gcp"
   collector_id  = "${sumologic_collector.collector.id}"
+  authentication {
+    type = "NoAuthentication"
+  }
+
+  path {
+    type = "NoPathExpression"
+  }
```
maimaisie, line 25: what happens if these are not set?

vsinghal13: updated the description

maimaisie: is that a bug in the code that it's not respecting defaults? also would the source not work (eg. not decode base64) when this happens?

frankreno (approved):
LGTM

maimaisie (approved):

yonghaoy:
Hi Just found this recently. Few questions:

1. Where can I find examples on how to use it? In our case, we need to setup some authentication to enable pubsub? There used to be some introduction https://help.sumologic.com/03Send-Data/Sources/02Sources-for-Hosted-Collectors/Google-Cloud-Platform-Source in this page tell us how to setup domain verification with GCP, but now it is removed. Not sure how to do it in Terraform
2. It may cause https://github.com/SumoLogic/terraform-provider-sumologic/issues/179 issue.

PR #136, opened 2021-01-15
Adding Role Data Source

Aayushi-Bansal:


sumologic/sumologic_role.go
```diff
@@ -53,6 +53,32 @@ func (s *Client) GetRole(id string) (*Role, error) {
...
+	err = json.Unmarshal(data, &response)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, r := range response.Roles {
+		if r.Name == name {
+			return &r, nil
+		}
+	}
```
seansain, line 77: The api has a query param called `name`. We should use that instead of going through all the roles.

Aayushi-Bansal: I searched for it earlier, just like we have for collectors, but wasn't able to find it.
Let me change it, if we really have the support of `name` query parameter in roles API as well.

seansain: Screen Shot 2021-01-15 at 9 27 17 AM
Should be the bottom one here

Aayushi-Bansal: Incorporated.

seansain (approved):

PR #154, opened 2021-02-11
Fix default value of analyticsTier for partitions

sumovishal:


sumologic/resource_sumologic_partition.go
```diff
@@ -33,7 +33,7 @@ func resourceSumologicPartition() *schema.Resource {
 				Type:         schema.TypeString,
 				Optional:     true,
 				ValidateFunc: validation.StringInSlice([]string{"continuous", "frequent", "infrequent", "enhanced", "basic", "cold"}, false),
```
seansain, line 35: Only new values are supported now right? We should get rid of the old ones.

seansain (approved):

PR #160, opened 2021-02-19
move length requirement to 0

seansain:


sumologic/resource_sumologic_partition.go
```diff
@@ -27,7 +27,7 @@ func resourceSumologicPartition() *schema.Resource {
 			"routing_expression": {
 				Type:         schema.TypeString,
 				Required:     true,
-				ValidateFunc: validation.StringLenBetween(1, 16384),
+				ValidateFunc: validation.StringLenBetween(0, 16384),
```
sumovishal, line 30: The API doesn't allow empty string for `routingExpression`. Does it matter if we remove it from here?

seansain: API was changed in September to allow updates of routing expressions. The problem on the API side is that you cannot send an empty routing expression, but you get one back.

sumovishal (approved):

PR #165, opened 2021-03-06
Fix schema for topology_label_map

sumovishal:
Terraform SDKv2 doesn't support map with non-primitive values. Workaround is to use a block object instead of a map.


emichaeli (approved):
LGTM

PR #170, opened 2021-03-12
Add trace source docs

frankreno:
Add trace source docs

### DIFF
--- a/website/docs/r/http_source.html.markdown
+++ b/website/docs/r/http_source.html.markdown
@@ -6,7 +6,7 @@ description: |-
 ---
 
 # sumologic_http_source
-Provides a [Sumologic HTTP source][1] or [Sumologic HTTP Traces source][2].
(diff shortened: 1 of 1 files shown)


frankreno:
@pmm-sumo @sumo-drosiek mind reviewing?

pmm-sumo (approved):

sumo-drosiek:
We should add note that tracing is not available for customers by default

website/docs/r/http_source.html.markdown
```diff
@@ -24,6 +24,14 @@ resource "sumologic_http_source" "http_source" {
   }
 }
 
+resource "sumologic_http_traces_source" "http_traces_source" {
```
vsinghal13, line 27: This should be `sumologic_http_source` resource itself.

frankreno: Good catch up dated the wrong property

frankreno: Fixed

vsinghal13 (approved):

PR #173, opened 2021-03-18
Read provider version info during compile time

sumovishal:
Use provider version info passed via `ldflags` instead of hard coded value. `goreleaser` passes in provider version via `ldflags` when making a new release.

### DIFF
--- a/sumologic/provider.go
+++ b/sumologic/provider.go
@@ -14,6 +14,7 @@ import (
 )
 
 func Provider() terraform.ResourceProvider {
+	log.Printf("Sumo Logic Terraform Provider Version=%s\n", ProviderVersion)
 	return &schema.Provider{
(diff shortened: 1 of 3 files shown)


sumologic/provider.go
```diff
@@ -14,6 +14,7 @@ import (
 )
 
 func Provider() terraform.ResourceProvider {
+	log.Printf("Sumo Logic Terraform Provider Version=%s\n", ProviderVersion)
```
seansain, line 17: do we still want the log line?

seansain (approved):

sumologic/provider.go
sumovishal: Yeah, I want that. This would help us identify the provider version when we ask customers for logs.

PR #177, opened 2021-03-30
Use int64 where API returns a 64 bit wide integer

pmalek-sumo:
Whenever the provider is built on a 32bit platform and a response from the API [1] is being unmarshalled, we get an error saying that an `int64` cannot be unmarshalled as `int`.

This comes from the fact that `Long` in Scala is a 64 bit wide integer and `int` in Go is platform dependent: being 32 bit wide integer on 32 bit platforms.

This PR addresses this problem in collector `struct` but there are other places in this repo where this needs addressing.

Thanks to @astencel-sumo for finding this!

[1]: https://help.sumologic.com/APIs/Collector-Management-API/Collector-API-Methods-and-Examples

### DIFF
--- a/sumologic/data_source_sumologic_collector.go
+++ b/sumologic/data_source_sumologic_collector.go
@@ -73,7 +73,7 @@ func dataSourceSumologicCollectorRead(d *schema.ResourceData, meta interface{}) 
 		}
 	}
 
-	d.SetId(strconv.Itoa(collector.ID))
+	d.SetId(strconv.FormatInt(collector.ID, 10))
 	d.Set("name", collector.Name)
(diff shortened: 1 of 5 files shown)


seansain:
Thanks @pmalek-sumo for addressing this! Do you know where else this needs to be addressed?

seansain (approved):

pmalek-sumo:
@seansain Not sure where. One would need to check the API responses against the resource definitions.

It should be fairly straightforward yet a bit mundane to do :)

e.g. here https://help.sumologic.com/APIs/Collector-Management-API/Collector-API-Methods-and-Examples#response-fields

PR #199, opened 2021-05-24
throw error for version key in schema_ref

vsinghal13:
Throw an error if `version` is specified in `schema_ref` map field in the cloud_to_cloud source.

### DIFF
--- a/sumologic/resource_sumologic_cloud_to_cloud_source.go
+++ b/sumologic/resource_sumologic_cloud_to_cloud_source.go
@@ -27,10 +27,18 @@ func resourceSumologicCloudToCloudSource() *schema.Resource {
 				DiffSuppressFunc: structure.SuppressJsonDiff,
 			},
 			"schema_ref": {
-				Type:     schema.TypeMap,
+				Type:     schema.TypeList,
 				Required: true,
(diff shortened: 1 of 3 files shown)


sumovishal:
Changing the type to `TypeList` from `TypeMap` is a breaking change. Why not just add a validation in the code to throw an appropriate error if a key other than `type` is present in the `schema_ref` object.

vsinghal13:
> Changing the type to `TypeList` from `TypeMap` is a breaking change. Why not just add a validation in the code to throw an appropriate error if a key other than `type` is present in the `schema_ref` object?

We discussed that but want to keep the behavior consistent with other sources. For eg. the authentication and path in polling sources also use the same concept.

sumovishal:
> We discussed that but want to keep the behavior consistent with other sources. For eg. the authentication and path in polling sources also use the same concept.

This will require a major version bump. It seems a bit silly to do a major version bump with only a minor change. I would vote to keep the current type and switch type to `List` later when we have other changes to go along with major version bump.

vsinghal13:
Changed to return an error if `version` key is specified in the `schema_ref` map.

maimaisie (approved):

PR #203, opened 2021-06-10
Add Terraform support for token

yuting-liu:
Add terraform support for installation tokens.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 2.9.4 (Unreleased)
 
-Features:
+FEATURES:
 
 * **New Resource:** sumologic_token (GH-203)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## 2.9.4 (Unreleased)
 
+Features:
+
+* **New Resource:** sumologic_token (GH-203)
+
 ## 2.9.3 (April 26, 2021)
 
 FEATURES:
--- a/sumologic/provider.go
+++ b/sumologic/provider.go
@@ -74,6 +74,7 @@ func Provider() terraform.ResourceProvider {
 			"sumologic_password_policy":                    resourceSumologicPasswordPolicy(),
 			"sumologic_saml_configuration":                 resourceSumologicSamlConfiguration(),
 			"sumologic_kinesis_metrics_source":             resourceSumologicKinesisMetricsSource(),
+			"sumologic_token":                              resourceSumologicToken(),
 		},
 		DataSourcesMap: map[string]*schema.Resource{
 			"sumologic_caller_identity": dataSourceSumologicCallerIdentity(),
--- a/sumologic/resource_sumologic_token.go
+++ b/sumologic/resource_sumologic_token.go
@@ -101,8 +101,7 @@ func resourceSumologicTokenUpdate(d *schema.ResourceData, meta interface{}) erro
 	if err != nil {
 		return err
 	}
-	// Each successful update will increment version
-	d.Set("version", token.Version+1)
+
 	return resourceSumologicTokenRead(d, meta)
 }
 
--- a/sumologic/resource_sumologic_token.go
+++ b/sumologic/resource_sumologic_token.go
@@ -4,6 +4,7 @@ import (
 	"log"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
 )
 
 func resourceSumologicToken() *schema.Resource {
@@ -39,8 +40,9 @@ func resourceSumologicToken() *schema.Resource {
 			},
 
 			"type": {
-				Type:     schema.TypeString,
-				Required: true,
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringInSlice([]string{"CollectorRegistration", "CollectorRegistrationTokenResponse"}, false),
 			},
 		},
 	}
--- a/sumologic/resource_sumologic_token.go
+++ b/sumologic/resource_sumologic_token.go
@@ -1,0 +1,118 @@
+package sumologic
+
+import (
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceSumologicToken() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceSumologicTokenCreate,
+		Read:   resourceSumologicTokenRead,
+		Update: resourceSumologicTokenUpdate,
+		Delete: resourceSumologicTokenDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"version": {
+				Type:     schema.TypeInt,
+				Optional: true,
+				Computed: true,
+			},
+
+			"status": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"name": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"type": {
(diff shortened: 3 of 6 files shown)


sumologic/resource_sumologic_token_test.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+)
+
+func TestAccSumologicToken_basic(t *testing.T) {
+	var token Token
+
+	testName := "Test Terraform Token"
+	testDescription := "Description tf test token"
+	testStatus := "Active"
```
vsinghal13, line 18: maybe we can use a function similar to `getRandomParams()` which is present in other test files, to set the values of these variables in all the tests.

sumologic/resource_sumologic_token.go
```diff
@@ -0,0 +1,117 @@
...
+			},
+
+			"description": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+
+			"type": {
+				Type:     schema.TypeString,
+				Required: true,
```
vsinghal13, line 43: can `type` have any other value apart from `CollectorRegistration`? If not, then we should add a check for the same.

vsinghal13 (commented):
Overall LGTM, pending some comments. 
Also, we should add relevant docs for the new resource.

sumologic/resource_sumologic_token.go
yuting-liu: Currently, we only have `CollectorRegistration` type for tokens.
But the tricky part is that the type will change between request and response.
In the request, the type for token is `CollectorRegistration`. However, in the response, type will be `CollectorRegistrationTokenResponse`. This is rendered by content library framework.

For example, the request body is:
```
{
  "name": "test-token",
  "description": "new token",
  "status": "Active",
  "type": "CollectorRegistration",
  "version": 0
}
```
And response will be like this:
```
{
    "id": "0000000000008A31",
    "name": "test-yuting",
    "description": "new token",
    "status": "Active",
    "type": "CollectorRegistrationTokenResponse",
    "version": 0,
    "createdAt": "2020-07-08T17:21:30.732Z",
    "createdBy": "0000000005C48CF8",
    "modifiedAt": "2021-06-10T17:10:43.489Z",
    "modifiedBy": "000000000B05888D",
    "encodedTokenAndUrl": "xxx"
}
```

sumologic/resource_sumologic_token_test.go
yuting-liu: @vsinghal13 I found the usage of method `getRandomizedParams()` in `resource_sumologic_cloud_to_cloud_source_test.go`. However, I didn't find its definition. How is this method working right now?

vsinghal13: https://github.com/SumoLogic/terraform-provider-sumologic/blob/master/sumologic/resource_sumologic_s3_source_test.go#L14

sumologic/resource_sumologic_token.go
yuting-liu: Had some discussions offline on the validation for type.
Decided to add type check for `CollectorRegistration` and `CollectorRegistrationTokenResponse`.

sumologic/provider.go
```diff
@@ -74,6 +74,7 @@ func Provider() terraform.ResourceProvider {
 			"sumologic_password_policy":                    resourceSumologicPasswordPolicy(),
 			"sumologic_saml_configuration":                 resourceSumologicSamlConfiguration(),
 			"sumologic_kinesis_metrics_source":             resourceSumologicKinesisMetricsSource(),
+			"sumologic_token":                              resourceSumologicToken(),
```
maimaisie, line 77: just one concern about the resource name, do we have any other kind of tokens in Sumo? should we name this `sumologic_collector_installation_token` or something more specific?

yuting-liu: This is a good call. 

Actually, this aligns with our api name

Screen Shot 2021-06-16 at 4 02 07 PM
 (https://nite-api.sumologic.net/docs/#tag/tokensLibraryManagement)

I think our original intention is that if there is any other kind of token in Sumo, we still want to have it use the same token, but change `type` to something else. Our current installation token is using the type of `CollectorRegistration`

vsinghal13 (approved):
LGTM, please add docs page as well.

yuting-liu:
@maimaisie @vsinghal13 @seansain ping for another round of review for tokens doc

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 ## 2.9.4 (Unreleased)
 
+Features:
```
maimaisie, line 3: needs to be all uppercase to stay consistent?

website/docs/r/token.html.markdown
```diff
@@ -0,0 +1,43 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+  * `name` - (Required) Display name of the token. This must be unique across all of the tokens.
+  * `description` - (Optional) The description of the token.
+  * `status` - (Required) Status of the token. Supported values are `Active`, and `Inactive`.
+  * `type` - (Required) Type of the token. Supported value is `CollectorRegistration`.
+  * `version` - (Required for update) Version of the token. This is only required for update. It starts from 0 when created and gets incremented with each update.
```
maimaisie, line 30: not TF specific question: why do we require user to manage the version?

CHANGELOG.md
yuting-liu: Fixed it.

website/docs/r/token.html.markdown
yuting-liu: As for `version`, this is required by content library framework.
It's not needed for `CREATE` since framework will have all objects with version starting from 0. The version is required for `UPDATE`. The reason is that we need to make sure they don't have inconsistent update. It works similarly to `Etag` in collector update request.

maimaisie: yeah I understand it's to prevent race conditions on update, yet we don't require user to specify an `etag` in terraform for sources do we?

yuting-liu: It's a good question. I dig into the code and it seems that it doesn't mention about the `ETag` anywhere in collector/source terraform. It seems that `ETag` is handled in [api.go](/Users/yuting/git/terraform-provider-sumologic/vendor/github.com/aws/aws-sdk-go/service/s3/api.go)

The current `required` for version aligns with our token [api doc](https://nite-api.sumologic.net/docs/#operation/updateToken). 

Want to have confirmation from TF SME on this @vsinghal13 @seansain

yuting-liu: Ping @vsinghal13 @seansain @sumovishal again on this issue.

seansain: From looking at the Monitors resource there is also a `version` parameter, but it is not documented. I will do some digging to check how it works.

sumologic/resource_sumologic_token.go
```diff
@@ -0,0 +1,119 @@
...
+func resourceSumologicTokenUpdate(d *schema.ResourceData, meta interface{}) error {
+	c := meta.(*Client)
+
+	token := resourceToToken(d)
+	err := c.UpdateToken(token)
+	if err != nil {
+		return err
+	}
+	// Each successful update will increment version
+	d.Set("version", token.Version+1)
```
seansain, line 105: I don't think this is needed. Follow same way that Monitors is handling `version`

website/docs/r/token.html.markdown
seansain: We can follow the same precedence as monitors here. We do not need to expose `version` in the docs. It is updated by the server and the right value is returned back to us so Terraform will store that. So, when the user tries to update the resource they will have the right `version`.

seansain (changes requested):

seansain (approved):

maimaisie (approved):
make sure to test create/update locally again now after the version is removed

yuting-liu:
Already verified that both create and update are working well.

PR #216, opened 2021-06-25
Fixing the sprintf formatting

karthich:
The last PR was my first time writing Go. My GoLand IDE made it look like I was doing string formatting correctly when I wrote these two lines. I was definitely doing it wrong. I would like to write a couple of tests against this pipeline before I am ready to merge it


karthich (commented):
Okay. I have the code changes with a test to confirm the admin recommended folder gets queried correctly. Since I can't really assert the ID I am just making sure its set and confirming the admin folder is called "Admin Recommended" which seems to be the default for sumologic.

seansain (approved):
I'll merge this and see if the tests pass. Thanks for fixing this.

PR #219, opened 2021-06-29
Misc fixes to content and folder resource

sumovishal:
1. Fixed `Status` and `Error` struct to correctly unmarshal API response.
2. Fixed content test to use correct auto parsing mode.
3. Cleaned up sumologic_folder, resource_sumologic_folder, resource_sumologic_content files.
4. Removed execute permissions from the files.


seansain (approved):
LGTM, might be worth moving to this [logger](https://github.com/sirupsen/logrus) in a future PR if we want to separate INFO, DEBUG, etc logs.

sumovishal:
> LGTM, might be worth moving to this [logger](https://github.com/sirupsen/logrus) in a future PR if we want to separate INFO, DEBUG, etc logs.

hmm, yeah, something like that will be useful. Though the project says it's in maintenance mode.

PR #220, opened 2021-06-29
Fixed timerange read

kumar-avijit:
* Fixed monitor read to set the time range from the monitor being read, after stripping the leading minus sign, if present.
* Fixed tests to verify time range.


sumologic/resource_sumologic_monitors_library_monitor_test.go
```diff
@@ -398,7 +401,7 @@ resource "sumologic_monitor" "test" {
 	triggers  {
 		threshold_type = "GreaterThan"
 		threshold = 40.0
-		time_range = "15m"
+		time_range = "30m"
```
rsanbhad, line 404: nit: we might want to have a testcase with negative timerange?

rsanbhad (approved):
lgtm

seansain (approved):

PR #230, opened 2021-07-09
Adding time range with negative terse value as allowed value in monitors

tarunk2:


rsanbhad (approved):

rsanbhad:
@seansain we will need a new provider released with this change ASAP (see thread in https://sumologic.slack.com/archives/C016AS0VDSA/p1625818466108700?thread_ts=1625756544.085300&cid=C016AS0VDSA )

PR #229, opened 2021-07-07
Added support for outlier and field-based monitors

kumar-avijit:
This adds support for 6 new monitor condition types:
* LogsStaticCondition
* MetricsStaticCondition
* LogsOutlierCondition
* MetricsOutlierCondition
* LogsMissingDataCondition
* MetricsMissingDataCondition

(@rsanbhad , @tarunk2 : I am still to fix some tests, but I want to get your early opinion on how justified the loss of strict schema checking for the sake of backwards compatibility is here.)

PR #233, opened 2021-07-09
Created sub-schemas for each monitor types

kumar-avijit:
A cleaner approach to adding new monitor types, than https://github.com/SumoLogic/terraform-provider-sumologic/pull/229

This provides strict schema-checking on `triggers` field for each of the monitor types, while still being backwards compatible.

### DIFF
--- a/sumologic/resource_sumologic_monitors_library_monitor.go
+++ b/sumologic/resource_sumologic_monitors_library_monitor.go
@@ -88,6 +88,71 @@ func resourceSumologicMonitorsLibraryMonitor() *schema.Resource {
 				Optional: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"trigger": {
+							Type: schema.TypeList,
+							MaxItems: 1,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"static_condition": {
+										Type: schema.TypeList,
+										MaxItems: 1,
+										Optional: true,
+										Elem: &schema.Resource{
+											Schema: getStaticConditionSchema(),
+										},
+									},
+									"logs_static_condition": {
+										Type: schema.TypeList,
+										MaxItems: 1,
+										Optional: true,
+										Elem: &schema.Resource{
+											Schema: getLogsStaticConditionSchema(),
+										},
+									},
+									"metrics_static_condition": {
+										Type: schema.TypeList,
+										MaxItems: 1,
+										Optional: true,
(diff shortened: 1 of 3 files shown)


sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -88,6 +88,71 @@ func resourceSumologicMonitorsLibraryMonitor() *schema.Resource {
 				Optional: true,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
+						"trigger": {
```
rsanbhad, line 91: This would be an explicit sub field of triggers if I understand this correctly? I don't think this would work with the current apis

sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -88,6 +88,71 @@ func resourceSumologicMonitorsLibraryMonitor() *schema.Resource {
...
+							MaxItems: 1,
+							Optional: true,
+							Elem: &schema.Resource{
+								Schema: map[string]*schema.Schema{
+									"static_condition": {
+										Type: schema.TypeList,
+										MaxItems: 1,
+										Optional: true,
+										Elem: &schema.Resource{
+											Schema: getStaticConditionSchema(),
```
rsanbhad, line 102: +1 to this paradigm even though it might cause significant code duplication

rsanbhad (commented):
I understand it might not be ready for review but I saw it referenced from another commit and took a look and lgtm so far

sumologic/resource_sumologic_monitors_library_monitor.go
kumar-avijit: Thanks for reviewing this! 
So this actually turns out to not be a concern since we eventually translate resource data from here to JSON model classes, which are then marshaled automatically to form the JSON payload) when making infrastructure calls from Terraform. When translating `trigger` blocks of the resource model to JSON, we flatten out the fields into `TriggerCondition`.

kumar-avijit: BTW, I am working on (yet) another, and a shorter version of this change. I realized that the `trigger` block is unnecessary here--we can put condition blocks directly under `triggers` like so
```
triggers {
  logs_outlier_condition {
   field = ...
   ...
  }
}
```
Will send it out for review shortly.

kumar-avijit:
superseded by https://github.com/SumoLogic/terraform-provider-sumologic/pull/239

PR #239, opened 2021-07-21
Support for new monitor types

kumar-avijit:
This PR implements support for new detection methods. 

Here's a summary of the changes on the HCL side:
1. A new top-level block `trigger_conditions`  is added and the old `triggers` block is marked as deprecated (but continues to be supported).
2. Each of the new detection methods appears as a sub-block inside `trigger_conditions`. Each of them comes with its own schema (see [here](https://github.com/SumoLogic/terraform-provider-sumologic/pull/239/files#diff-3b5cf55c9167896c315794991e5270d17c43dfe94d285b2fcf257ad9e4a76d03R588-R649) for examples).

On the JSON side:
1. The `TriggerCondition` object is extended with additional fields, such as `baseline_window` etc., that are required by the new detection methods. 

Note that this means that at the JSON level, the trigger condition is represented in a _flattened_ form, in contrast to the nested form in HCL. This ensures compatibility with existing infrastructure APIs.

Here's how a typical `triggers` configuration might look like:
```
  logs_static_condition {
    field = "_count"
    critical {
      time_range = "15m"
      alert {
        threshold = 100
        threshold_type = "GreaterThan"
      }
      resolution {
        threshold = 90
        threshold_type = "LessThanOrEqual"
      }
    }
    warning {
      time_range = "30m"
      alert {
        threshold = 80
        threshold_type = "GreaterThan"
      }
      resolution {
        threshold = 75
        threshold_type = "LessThanOrEqual"
      }
    }
  }
  logs_missing_data_condition {
    time_range = "30m"
  }
}
```

#### Suggested review plan
LOC-wise, this is quite a big PR. The bulk of changes are in the `resource_sumo~_monitor.go` file. Here you'll find:
1.  schema definitions for `trigger_conditions` and its subblocks (please review these)
2. code that deals with conversions between the object representing `trigger_conditions` block in TF state and the `TriggerCondition` class, which represents the JSON side of things.
    ```
    func jsonToTriggerConditionsBlock(conditions []TriggerCondition) map[string]interface{} 
    ```
    and
    ```
    func triggerConditionsBlockToJson(block map[string]interface{}) []TriggerCondition
    ```
    Please also review these methods, but maybe go no more than 1 level deeper (beyond the individual sub-methods) into each of these. 
3. A bunch of code, mostly relegated toward the end of file, that deals with misc. utility-like functionality. You might want to just skim this, if not wholly ignore it. 

As for the rest of the files, the code changes are quite small.

#### Testing
- [x] Examples in docs work.
- [x] Unit tests and acceptance tests pass.

#### Feedback from 7/21 review meeting:
- [x] Test whether updating a `StaticCondition` monitor to a `LogsStaticCondition` monitor in TF works
     - [x] Create StaticCondition monitor using `triggers` block with v2.9.6 > `tf init` to new version > Verify `tf plan + apply` works > Translate monitor to use `trigger_conditions.logs_static_condition` > Verify `tf plan + apply` works. 
- [x] Remove `StaticCondition` from `trigger_conditions`
- [x] Move `time_range` to inside `critical` and `warning` blocks + any other fields that might need moving
- [ ] Can we mark / pull bad provider versions from TF repo?
- [ ] Research syntax for list type blocks


sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -358,21 +573,35 @@ func resourceSumologicMonitorsLibraryMonitorRead(d *schema.ResourceData, meta in
...
-			"trigger_type":     t.TriggerType,
-			"threshold":        t.Threshold,
-			"threshold_type":   t.ThresholdType,
-			"time_range":       strings.TrimPrefix(t.TimeRange, "-"),
-			"occurrence_type":  t.OccurrenceType,
-			"trigger_source":   t.TriggerSource,
-			"detection_method": t.DetectionMethod,
+
+	// set either 'trigger_conditions' or 'triggers', but not both, based on whichever the plan uses.
+	// we avoid converting between the 2 so as to prevent plan mismatches before and after an apply.
```
tarunk2, line 578: nit: between the two

tarunk2 (approved):
LGTM, good job!

sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -470,23 +699,370 @@ func getNotifications(d *schema.ResourceData) []MonitorNotification {
...
+			}
+		}
+		return triggers
+	}
+}
+
+// 'trigger_conditions' block to TriggerCondition JSON structure
+func triggerConditionsBlockToJson(block map[string]interface{}) []TriggerCondition {
+	conditions := make([]TriggerCondition, 0)
+	if sc, ok := fromSingletonArray(block, logsStaticConditionFieldName); ok {
```
rsanbhad, line 727: nit: do we need error handling for the `ok` values here? (I think the `ok` is error code in this context right?)

sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -256,6 +315,162 @@ func resourceSumologicMonitorsLibraryMonitor() *schema.Resource {
...
+		"baseline_window": &baselineWindowSchema,
+		"threshold":       &thresholdSchema,
+	}),
+}
+
+var logsMissingDataTriggerConditionSchema = map[string]*schema.Schema{
+	"time_range": {
+		Type:         schema.TypeString,
+		Required:     true,
+		ValidateFunc: validation.StringInSlice([]string{"5m", "-5m", "10m", "-10m", "15m", "-15m", "30m", "-30m", "60m", "-60m", "1h", "-1h", "3h", "-3h", "6h", "-6h", "12h", "-12h", "24h", "-24h", "1d", "-1d"}, false),
```
rsanbhad, line 417: nit: I was thinking there should be a way to define a custom `ValidationFunc` (as a subtype of `SchemaValidationFunc` I guess? https://pkg.go.dev/github.com/hashicorp/terraform-plugin-sdk@v1.17.2/helper/schema#SchemaValidateFunc ) separately and then re-use it across the various trigger conditions for fields where the validation will be common/shared (like timerange) - but I couldn't find an obvious way with a quick search 🤔  either way shouldn't really matter too much

sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -126,6 +127,64 @@ func resourceSumologicMonitorsLibraryMonitor() *schema.Resource {
 				},
 			},
 
+			"trigger_conditions": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
```
rsanbhad, line 134: thanks for breaking down the schema! this is pretty clean 😄

sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -534,3 +1110,149 @@ func resourceToMonitorsLibraryMonitor(d *schema.ResourceData) MonitorsLibraryMon
...
+		return &schema.Schema{
+			Type:     schema.TypeList,
+			Required: true,
+			MaxItems: 1,
+			Elem:     toResource(sch),
+		}
+	}
+}
+
+func (condition *TriggerCondition) readFrom(block map[string]interface{}) {
```
rsanbhad, line 1147: oh this is pretty neat, I didn't know we could case match like this 🤔

rsanbhad (approved):
lgtm, thanks!

rsanbhad:
One comment I forgot, if you get a chance to update the CHANGELOG.md with the release notes for this change, I think it makes it easier for terraform team/Sean when it comes time to release the new provider version

sumologic/resource_sumologic_monitors_library_monitor.go
kumar-avijit: Ah, thanks for pointing out. Looks like I forgot to replace the block with  
```
"time_range":      &timeRangeSchema,
```
here (and in the next schema), like I did for the rest. Will fix.

sumologic/resource_sumologic_monitors_library_monitor.go
kumar-avijit: Maybe, but not super necessary, imo.

The `ok` variable (which is of type `bool`, not an `error`) could be `false` for a field for valid reasons, e.g., when the block does not contain that field. 

That said, a bug in the provider logic could also lead to `ok` being `false` when this function is called on a state, e.g., when the block contains the field, but its value is not in the expected form. Maybe in that case it'll be better to have `fromSingletonArray` panic instead of silently returning `..., false`? 🤔

kumar-avijit: Ended up making `fromSingletonArray` more robust against bad data.

sumologic/resource_sumologic_monitors_library_monitor.go
kumar-avijit: fixed.

PR #241, opened 2021-07-21
update go version

### LINKED ISSUE #226: Add arm64 support for Apple silicon
A lot of engineers building infrastructure use Apple MacBooks/Macs.  Latest gen machines run the arm64-based M1 chip.  Please consider adding support for this architecture.  

We cannot run Terraform code that uses the Sumologic provider which severely hinders local development.

seansain:
I think this might fix #226


sumovishal (approved):
🤞🏽

PR #246, opened 2021-07-22
remove mutexkv import

seansain:
`mutexkv` is removed from v2 of the terraform plugin sdk
I don't think this lock was providing anything either since we wouldn't be modifying the `urlpath` elsewhere while doing the put.

### DIFF
--- a/sumologic/sumologic_client.go
+++ b/sumologic/sumologic_client.go
@@ -186,9 +186,6 @@ func (s *Client) PostRawPayload(urlPath string, payload string) ([]byte, error) 
 }
 
 func (s *Client) Put(urlPath string, payload interface{}, isAdminMode bool) ([]byte, error) {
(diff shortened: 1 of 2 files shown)


sumologic/sumologic_client.go
```diff
@@ -186,9 +186,6 @@ func (s *Client) PostRawPayload(urlPath string, payload string) ([]byte, error)
 }
 
 func (s *Client) Put(urlPath string, payload interface{}, isAdminMode bool) ([]byte, error) {
-	SumoMutexKV.Lock(urlPath)
```
sumovishal, line 189: This was serializing update calls to a resource. Now we may process 2 update requests at the same time. This may be okay provided we process the requests appropriately in the backend. We should keep an eye out for issue around update calls.

sumovishal (approved):

PR #248, opened 2021-07-22
Add Policies Resource

emichaeli:
Add Policies resource


website/docs/r/policies.html.markdown
```diff
@@ -47,6 +47,12 @@ The following arguments are supported:
 - `enabled` - (Optional) Whether the [User Concurrent Sessions Limit Policy][6] is enabled. Defaults to `false`.
 - `max_concurrent_sessions` - (Optional) Maximum number of concurrent sessions a user may have. Defaults to `100`.
 
+## Import
+Policies can be imported using the id `org-policies`, e.g.:
```
sumovishal, line 51: We should make it clear here that only `org-policies` would work, and specifying some other identifier will have unintended consequences. Use the following text for a warning block.

```
~> Warning text
```

sumovishal (approved):
lgtm, one minor comment for the doc part.

PR #260, opened 2021-08-06
Add Terraform support for hierarchy

mfiglus:
Add Terraform support for entities hierarchies

Hierarchies API uses recursively defined structures, as a workaround TF schemas are defined through those recursions expanded to some set depth.

### DIFF
--- a/sumologic/resource_sumologic_hierarchy.go
+++ b/sumologic/resource_sumologic_hierarchy.go
@@ -1,0 +1,326 @@
+package sumologic
+
+import (
+	"log"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+)
+
+var (
+	maxDepth            = 6
+	maxWidth            = 20
+	maxHierarchyNameLen = 256
+	maxFilterKeyLen     = 1024
+	maxFilterValueLen   = 1024
+	maxEntityTypeLen    = 1024
+)
+
+func resourceSumologicHierarchy() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceSumologicHierarchyCreate,
+		Read:   resourceSumologicHierarchyRead,
+		Update: resourceSumologicHierarchyUpdate,
+		Delete: resourceSumologicHierarchyDelete,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"name": {
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringLenBetween(1, maxHierarchyNameLen),
+			},
+
+			"filter": {
+				Type:     schema.TypeList,
+				Optional: true,
+				MaxItems: 1,
+				Elem: &schema.Resource{
+					Schema: getFilterSchema(),
+				},
+			},
+
+			"level": {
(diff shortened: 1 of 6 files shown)


website/docs/r/hierarchy.html.markdown
```diff
@@ -0,0 +1,62 @@
...
+  
+The following attributes are exported:
+
+- `id` - The internal ID of the hierarchy.
+
+## Import
+Hierarchies can be imported using the name, e.g.:
+
+```hcl
+terraform import sumologic_hierarchy.test id
```
mccartney, line 59: > imported using the name

The example mentions `id` and not `name`. I'd suggest to clarify.

mfiglus:
Test failures seem to be unrelated to the PR

sumologic/resource_sumologic_hierarchy.go
```diff
@@ -0,0 +1,326 @@
...
+					Type: schema.TypeString,
+				},
+			},
+		}
+	}
+	for k, v := range nextLevelsWithConditionsSchema {
+		levelSchema[k] = v
+	}
+
+	if currentDepth < maxDepth {
```
sumovishal, line 108: What is the behavior when depth > maxDepth? Do we silently drop levels after max depth is reached?

website/docs/r/hierarchy.html.markdown
```diff
@@ -0,0 +1,60 @@
...
+
+## Argument Reference
+
+The following arguments are supported:
+
+- `name` - (Required) Name of the hierarchy.
+- `filter` - (Optional) An optional clause that a hierarchy requires to be matched.
+  + `key` - (Required) Filtering key.
+  + `value` - (Required) Value required for the filtering key.
+- `level` - (Required) A hierarchy of entities. The order is up-down, left to right levels with condition, then level without condition.
```
sumovishal, line 42: Can we document max depth we support for levels here?

sumovishal (approved):
lgtm, left some minor comments. Please address them before merging.

sumologic/resource_sumologic_hierarchy.go
```diff
@@ -0,0 +1,326 @@
...
+		},
+	}
+}
+
+func getFilterSchema() map[string]*schema.Schema {
+	return map[string]*schema.Schema{
+		"key": {
+			Type:         schema.TypeString,
+			Required:     true,
+			ValidateFunc: validation.StringLenBetween(1, maxFilterKeyLen),
```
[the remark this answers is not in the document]
mfiglus: They may not be present in API definition itself but we do have them in the hierarchy validation code

sumologic/resource_sumologic_hierarchy.go
mfiglus: If I understand correctly the provider will not accept such hierarchy because of unexpected field being present

sumovishal: I think what you said makes sense. Can you test this locally by setting `maxDepth` to a lower value, like 2?

mfiglus:
```
- Unsupported block type: Blocks of type "next_levels_with_conditions" are not expected here. Did you mean to define argument "next_levels_with_conditions"? If so, use the equals sign to assign it a value.
- Unsupported block type: Blocks of type "next_level" are not expected here.
```

sumovishal:
I am going to merge the PR, given test failure is unrelated. Sean is working on fixing the failing test.

PR #261, opened 2021-08-06
Reduced nesting level to fix formatting issues

kumar-avijit:
Terraform seems to have messed up indentation in arguments reference. The published [docs](https://registry.terraform.io/providers/SumoLogic/sumologic/latest/docs/resources/monitor):
image
instead of 
image

I suspect if this was caused as a result of the nesting depth in that argument list being one level too deep for Terraform? 
This PR reduces the nesting depth. Hopefully this will fix formatting and will come out like so:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ BUG FIXES:
 
 * resource/sumologic_monitor: Removed deprecation warning for `triggers`.
 * seperated docs for sumologic_monitor_folder from docs for sumologic_monitor.
+* resource/sumologic_monitor: Fixed docs for `trigger_conditions`.
 
 ## 2.9.8 (July 30, 2021)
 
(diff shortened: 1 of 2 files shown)


rsanbhad (approved):
lgtm overall - I think there's a way to test the website locally (`make web`?) but IIRC I don't think i ever got that working in the past 🤔  something to try if you get a chance, but otherwise it should still be fine

kumar-avijit:
> lgtm overall - I think there's a way to test the website locally (`make web`?) but IIRC I don't think i ever got that working in the past 🤔 something to try if you get a chance, but otherwise it should still be fine

Ah, that's good to know. 

Played around with `make website`, but couldn't get it working either; got it to a point where the web server was up and running, but http://localhost:4567/docs/providers/sumologic threw a 404. Traced it to ~/go/src/github.com/hashicorp/terraform-website/ext/terraform` being unpopulated, and then gave up.

sumovishal:
> lgtm overall - I think there's a way to test the website locally (`make web`?) but IIRC I don't think i ever got that working in the past 🤔 something to try if you get a chance, but otherwise it should still be fine

I use Doc Preview Tool (https://registry.terraform.io/tools/doc-preview) when making doc changes. Works like charm.

PR #263, opened 2021-08-11
Sumo 169842 terraform delay field

ndhar-sumo:
Added new field 'evaluation_delay' to terraform schema

### DIFF
--- a/sumologic/resource_sumologic_monitors_library_monitor.go
+++ b/sumologic/resource_sumologic_monitors_library_monitor.go
@@ -268,9 +268,9 @@ func resourceSumologicMonitorsLibraryMonitor() *schema.Resource {
 			},
 
 			"evaluation_delay": {
-			  Type:     schema.TypeString,
-      	Optional: true,
-			}
+				Type:     schema.TypeString,
+				Optional: true,
+			},
 
 			"is_locked": {
 				Type:     schema.TypeBool,
@@ -1090,7 +1090,7 @@ func resourceToMonitorsLibraryMonitor(d *schema.ResourceData) MonitorsLibraryMon
 		CreatedAt:          d.Get("created_at").(string),
 		MonitorType:        d.Get("monitor_type").(string),
 		Description:        d.Get("description").(string),
-		EvaluationDelay     d.Get("evaluation_delay").(string),
+		EvaluationDelay:    d.Get("evaluation_delay").(string),
 		Queries:            queries,
 		ModifiedBy:         d.Get("modified_by").(string),
 		IsMutable:          d.Get("is_mutable").(bool),
--- a/sumologic/resource_sumologic_monitors_library_monitor.go
+++ b/sumologic/resource_sumologic_monitors_library_monitor.go
@@ -272,7 +272,7 @@ func resourceSumologicMonitorsLibraryMonitor() *schema.Resource {
 				Type:         schema.TypeString,
 				Optional:     true,
 				Computed:     true,
-				ValidateFunc: validation.StringMatch(regexp.MustCompile(`((\d)+[smh])+`), "this value is bad"),
+				ValidateFunc: validation.StringMatch(regexp.MustCompile(`((\d)+[smh])+`), "This value is not in correct format. Example: 1m30s"),
 			},
 
 			"is_locked": {
--- a/sumologic/resource_sumologic_monitors_library_monitor.go
+++ b/sumologic/resource_sumologic_monitors_library_monitor.go
@@ -512,7 +512,6 @@ func resourceSumologicMonitorsLibraryMonitorRead(d *schema.ResourceData, meta in
 	d.Set("created_by", monitor.CreatedBy)
 	d.Set("created_at", monitor.CreatedAt)
 	d.Set("monitor_type", monitor.MonitorType)
-	log.Printf("[WARN] value of delay: %v", monitor.EvaluationDelay)
 	d.Set("evaluation_delay", monitor.EvaluationDelay)
 	d.Set("modified_by", monitor.ModifiedBy)
 	d.Set("is_mutable", monitor.IsMutable)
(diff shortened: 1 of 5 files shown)


sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -267,6 +267,11 @@ func resourceSumologicMonitorsLibraryMonitor() *schema.Resource {
 				ValidateFunc: validation.StringInSlice([]string{"Logs", "Metrics"}, false),
 			},
 
+			"evaluation_delay": {
```
rsanbhad, line 270: if this will be returned in the response from the server by default, you might also want to put `Computed: true` here so the terraform state doesn't show a diff after apply - but if you get a chance you should try it out first and see if it's needed

sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -504,6 +509,7 @@ func resourceSumologicMonitorsLibraryMonitorRead(d *schema.ResourceData, meta in
 	d.Set("created_by", monitor.CreatedBy)
 	d.Set("created_at", monitor.CreatedAt)
 	d.Set("monitor_type", monitor.MonitorType)
+	d.Set("evaluation_delay", monitor.EvaluationDelay)
```
rsanbhad, line 512: you might want to do something similar to what we do for timerange by calling `PositiveTimeRange` function

rsanbhad (approved):
some comments, overall lgtm - if you get a chance, try to test locally against nite (separate from the acceptance tests/UTs) as I find it helps catch more bugs and edge cases

sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -267,6 +267,11 @@ func resourceSumologicMonitorsLibraryMonitor() *schema.Resource {
 				ValidateFunc: validation.StringInSlice([]string{"Logs", "Metrics"}, false),
 			},
 
+			"evaluation_delay": {
+				Type:     schema.TypeString,
+				Optional: true,
```
kumar-avijit, line 272: You might want to add a `ValidateFunc` here for input validation. See `time_range` for an example.

kumar-avijit: I realize you might not be able to just use `time_range`'s validation function here as you probably want to admit a broader set of values. 

If that's the case, see if you could express it as a regex. You could then hand-code a validation function as done [here](https://www.terraform.io/docs/extend/schemas/schema-behaviors.html#validatefunc).

kumar-avijit (approved):
one comment, but lgtm.

sumologic/resource_sumologic_monitors_library_monitor.go
ndhar-sumo: I ran a test where i created a monitor on longData (prod dep). It created successfully. What diff was expected?
See below output:
`sumologic_monitor.test_log_static_resultcount: Creating...
sumologic_monitor.test_log_static_resultcount: Creation complete after 2s [id=000000000000ADAB]`

sumologic/resource_sumologic_monitors_library_monitor.go
ndhar-sumo: For this field, "0s" is allowed - that is not positive correct?

sumologic/resource_sumologic_monitors_library_monitor.go
ndhar-sumo: There is lot of validation code (Monitor manager validators) for the delay values permitted. When we try to create a monitor using terraform, won't that validation code return error messages to user? cc @rsanbhad

ndhar-sumo: or maybe no, because those are REST API messages?

sumologic/resource_sumologic_monitors_library_monitor.go
ndhar-sumo: Did you mean that even if i don't specify the field in request, does the response always have it?

sumologic/resource_sumologic_monitors_library_monitor.go
rsanbhad: In this context, positive just means without the negative sign (if you look at the implementation of PositiveTimeRange function you can see it just strips the leading "-" minus sign, if it exists)

sumologic/resource_sumologic_monitors_library_monitor.go
rsanbhad: if you manually create the monitor with this version of the provider and try to do a `terraform apply` after a successful terraform apply, does terraform say there are changes to apply, or does it say no changes detected? if no changes detected then you're fine

sumologic/resource_sumologic_monitors_library_monitor.go
rsanbhad: those error messages will happen on `terraform apply`, but I think Avijit is asking if we can do similar validations here - if you validate in the terraform provider, then those validations run on `terraform plan` which is a better user experience (it was raised in the past that monitors terraform could do more validation locally rather than relying on server validation errors, as those make it more difficult for users to know if their terraform configuration is correct when writing it)

sumologic/resource_sumologic_monitors_library_monitor.go
ndhar-sumo: I added "Computed: true" and also there are no changes to apply once the monitor has been created with the 1st `terraform apply`

sumologic/resource_sumologic_monitors_library_monitor.go
ndhar-sumo: i have added ValidateFunc

kumar-avijit: nice!

PR #265, opened 2021-08-30
set authentication block in polling sources

vsinghal13:


maimaisie:
I wonder if not setting the authentication block was done on purpose for security reasons? The credentials would be written in plaintext in the state file if it's set right?

vsinghal13:
> I wonder if not setting the authentication block was done on purpose for security reasons? The credentials would be written in plaintext in the state file if it's set right?

Ah, right. @sumovishal what do you suggest in this scenario, then?

Issue #115, opened 2020-11-26
Feature request - ability to manage Admin Recommended content


mccartney:
I'd be great to be able to mark some content as Admin Recommended in Terraform using this feature: https://help.sumologic.com/Manage/Content_Sharing/Content_Sharing_FAQ#i-want-to-create-a-hierarchy-of-folders-for-my-company-so-that-e

mccartney:
Related to #215 #211

sumovishal:
@mccartney this should be resolved now? We have AdminRecommended data source now.

mccartney:
Related to #322

PR #276, opened 2021-09-19
Enable admin mode in provider

kornypoet:
Set adminMode header in provider/client, to clean up http logic and allow for operations to optionally require it.
Addresses SumoLogic/terraform-provider-sumologic#115

### DIFF
--- a/sumologic/provider.go
+++ b/sumologic/provider.go
@@ -36,7 +36,7 @@ func Provider() terraform.ResourceProvider {
 				Optional: true,
 				Default:  os.Getenv("SUMOLOGIC_BASE_URL"),
 			},
-			"admin": {
+			"admin_mode": {
 				Type:     schema.TypeBool,
 				Optional: true,
 				Default:  false,
@@ -123,7 +123,7 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	accessKey := d.Get("access_key").(string)
 	environment := d.Get("environment").(string)
 	baseUrl := d.Get("base_url").(string)
-	isAdmin := d.Get("admin").(bool)
+	isInAdminMode := d.Get("admin_mode").(bool)
 
 	msg := ""
 	if accessId == "" {
@@ -158,6 +158,6 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 		accessKey,
 		environment,
 		baseUrl,
-		isAdmin,
+		isInAdminMode,
 	)
 }
--- a/sumologic/provider.go
+++ b/sumologic/provider.go
@@ -36,6 +36,11 @@ func Provider() terraform.ResourceProvider {
 				Optional: true,
 				Default:  os.Getenv("SUMOLOGIC_BASE_URL"),
 			},
+			"admin": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  false,
+			},
 		},
 		ResourcesMap: map[string]*schema.Resource{
 			"sumologic_cse_network_block":                  resourceSumologicCSENetworkBlock(),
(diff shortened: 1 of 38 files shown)


sumologic/provider.go
```diff
@@ -36,6 +36,11 @@ func Provider() terraform.ResourceProvider {
 				Optional: true,
 				Default:  os.Getenv("SUMOLOGIC_BASE_URL"),
 			},
+			"admin": {
```
sumovishal, line 39: I don't think is the right place for setting adminMode. The admin mode is only applicable to certain resources and should not be a global property. Besides having adminMode here will turn it on or off for all the resources. We want to toggle admin mode for each resource rather than globally.

If the goal is to get rid of adminMode and simplify call sites, we should look into some other option. Maybe have 2 versions of Get/Post/Put methods like GetInAdminMode/PostInAdminMode.

travis-jask: The pattern I followed here is similar to configuring an AWS provider to spin up resources in a different region; create a second provider and alias it. The way this would work in practice with sumologic would be:

```
provider "sumologic" {
  alias    = "admin"
  admin    = true
  base_url = "https://long-api.sumologic.net/api/"
}
```

Which would only be used in corresponding resources that need/want the admin call:

```# data "sumologic_admin_recommended_folder" "admin" {}

resource "sumologic_folder" "folder" {
  provider    = sumologic.admin
  name        = "test-folder"
  description = "A test folder"
  parent_id   = data.sumologic_admin_recommended_folder.admin.id
}
```

sumovishal: Ah, I see, I wasn't aware of this pattern. Yeah, this would work. However, this will a breaking change, as earlier we used to default to true for the admin recommended folder.

sumovishal: We should call this field "adminMode".

sumologic/provider.go
```diff
@@ -117,6 +122,7 @@ func providerConfigure(d *schema.ResourceData) (interface{}, error) {
 	accessKey := d.Get("access_key").(string)
 	environment := d.Get("environment").(string)
 	baseUrl := d.Get("base_url").(string)
+	isAdmin := d.Get("admin").(bool)
```
sumovishal, line 125: We should rename this to "isInAdminMode" as this only pertains to admin mode.

sumologic/provider.go
travis-jask: Is there a test that should pass/fail with this change? They pass locally for me. Additionally, the data source for `sumologic_admin_recommended_folder` seems to still work as well. The API docs don't require the header to get the folder ID: https://api.sumologic.com/docs/#operation/getAdminRecommendedFolderAsync ; it's only required when creating folders underneath it.

sumovishal (commented):
lgtm, can you update the changelog as well?

sumologic/provider.go
sumovishal: >The API docs don't require the header to get the folder ID: api.sumologic.com/docs/#operation/getAdminRecommendedFolderAsync ; it's only required when creating folders underneath it.

I thought we did, but looks like we don't. This will not be a breaking change in that case. 

>Is there a test that should pass/fail with this change?

There is https://github.com/SumoLogic/terraform-provider-sumologic/blob/master/sumologic/data_source_sumologic_admin_recommended_folder_test.go

sumovishal (approved):

sumovishal:
You will need to pull the latest master to fix build errors
```
sumologic/sumologic_cse_network_block.go:9:23: too many arguments in call to s.Get
```

travis-jask:
Rebased on master

sumovishal:
The tests are failing b/c access key/id are not set. Will merge the PR and run the test workflow in the master branch.

sumovishal:
> The tests are failing b/c access key/id are not set. Will merge the PR and run the test workflow in the master branch.

https://github.com/SumoLogic/terraform-provider-sumologic/actions/runs/1263936401

PR #281, opened 2021-09-23
Add support for CSE rule tuning expression

pmontiel-sumo:
Add support for CSE rule tuning expression

PR #285, opened 2021-09-28
Address missing comments from GH-274 and GH-281

pmontiel-sumo:
Address missing comments from GH-274 and GH-281

### DIFF
--- a/sumologic/sumologic_client.go
+++ b/sumologic/sumologic_client.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
-	"log"
 	"net/http"
 	"net/url"
 	"time"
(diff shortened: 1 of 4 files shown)


sumologic/sumologic_client.go
```diff
@@ -132,8 +134,10 @@ func (s *Client) GetWithCookies(urlPath string, cookies []*http.Cookie) ([]byte,
 func (s *Client) Post(urlPath string, payload interface{}) ([]byte, error) {
 	relativeURL, _ := url.Parse(urlPath)
 	sumoURL := s.BaseURL.ResolveReference(relativeURL)
-
+	log.Printf("[INFO] Post URL: %s", sumoURL)
```
sumovishal, line 137: Do we need these log lines? If yes, maybe they should be at DEBUG level?

sumovishal (approved):

sumologic/sumologic_client.go
pmontiel-sumo: Sorry my bad, I should have not commit that file; used those in my local testing, sorry
rollback that under https://github.com/SumoLogic/terraform-provider-sumologic/pull/285/commits/0d575a991ddbda1283d623aaa6b8975fac0067e7

PR #295, opened 2021-10-21
Add common Source properties

vsinghal13:


sumovishal:
> @seansain @sumovishal some tests related to `TestAccSumologicUser` are failing due to an existing role name. Could you please take a look?

I have merged the PR. Will look into the test failures (they are unrelated).

PR #301, opened 2021-10-29
SUMO-172016 add JWT support

dagould:


sumologic/provider.go
```diff
@@ -134,22 +143,37 @@ func resolveRedirectURL(accessId string, accessKey string) (string, error) {
...
 	msg := ""
-	if accessId == "" {
-		msg = "sumologic provider: access_id should be set;"
-	}
-
-	if accessKey == "" {
-		msg = fmt.Sprintf("%s access_key should be set; ", msg)
+	if authJwt == "" {
+		if accessId == "" || accessKey == "" {
+			msg = "sumologic provider: auth_jwt is not set;"
```
emichaeli, line 154: Since we are not documenting JWT authentication at this point and only using it internally, if both auth JWT and access id/key are not specified, we probably do not want to display `auth_jwt is not set;` in the error message. The most likely scenerio is that a user forgot to specify the access id/key, and the proper response is that the access id / key is missing. I.e. IMO we should not expose `auth_jwt` in any public response unless 100% sure that the user intended to use JWT auth.

sumologic/provider.go
```diff
@@ -134,22 +143,37 @@ func resolveRedirectURL(accessId string, accessKey string) (string, error) {
...
+	} else {
+		if accessId != "" || accessKey != "" {
+			msg = "sumologic provider: auth_jwt is set;"
+		}
+		if accessId != "" {
+			msg = fmt.Sprintf("%s access_id should not be set;", msg)
+		}
+		if accessKey != "" {
+			msg = fmt.Sprintf("%s access_key should not be set; ", msg)
+		}
```
emichaeli, line 171: I don't think that having access id / key + auth_jwt in the env variables should necessarily warrant an error response. For example, we might have both of them set in the environment and want to alternate between using JWT auth or the access id/key auth depending on the operation intended.

emichaeli (commented):
I think that since we are not documenting JWT auth publicly at this point and only using it internally, we should be very careful about not having it show up in the error response / logs when customers are using our provider.

We might want to think a little more about how we want to add this to the provider. I would personally suggest adding another env variable, e.g. `USE_JWT_AUTH`, and have your logic apply only if that env variable is set. I.e. if that env variable is not set, no errors / logs related to JWT auth should show up. Also, that way a user can set both access id/key auth and jwt auth, and alternate between which auth to use based on the use case.

sumologic/provider.go
```diff
@@ -18,14 +18,19 @@ func Provider() terraform.ResourceProvider {
...
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
 				DefaultFunc: schema.EnvDefaultFunc("SUMOLOGIC_ACCESSKEY", nil),
 			},
+			"auth_jwt": {
+				Type:        schema.TypeString,
+				Optional:    true,
+				DefaultFunc: schema.EnvDefaultFunc("SUMOLOGIC_AUTHJWT", nil),
+			},
```
emichaeli, line 33: Since we're only using jwt auth internally at the moment, I'm not sure if we even want/need to add it to the schema. Maybe just having it set as env variable only (and retrieving it from `os.Getenv` in the code) is enough.

sumologic/provider.go
```diff
@@ -18,14 +18,19 @@ func Provider() terraform.ResourceProvider {
 		Schema: map[string]*schema.Schema{
 			"access_id": {
 				Type:        schema.TypeString,
-				Required:    true,
+				Optional:    true,
```
sumovishal, line 21: I am not sure if  we want to make them optional. This is still our primary way to authenticate. We can set to them to some dummy string for our use case.

sumologic/provider.go
sumovishal: +1

sumologic/provider.go
sumovishal: +1. We should not expose JWT auth  in error messages.

sumologic/provider.go
sumovishal, line 171: Given that JWT auth is for our internal use case, should we ignore accessId/Key if JWT  auth is set?

dagould: Yes, with changing access id/key back to `Required`, this section is removed.

sumologic/provider.go
dagould: Ok, changing them back to `Required`, and changing the checking logic accordingly

sumologic/provider.go
dagould: Done.

sumologic/provider.go
dagould: Right -- removed this whole section, in accordance with changing access id/key back to `Required` (with dummy values to be passed in the JWT case). The logic is now: if JWT is not provided, have the old checks/messages for access id/key; and if it is, ignore them.

sumologic/provider.go
dagould: Done.

sumologic/provider.go
```diff
@@ -134,22 +138,27 @@ func resolveRedirectURL(accessId string, accessKey string) (string, error) {
...
 	msg := ""
-	if accessId == "" {
-		msg = "sumologic provider: access_id should be set;"
-	}
-
-	if accessKey == "" {
-		msg = fmt.Sprintf("%s access_key should be set; ", msg)
+	if authJwt == "" {
+		if accessId == "" || accessKey == "" {
+			msg = "sumologic provider: "
```
emichaeli, line 149: nit: you should remove the trailing space here as otherwise the resulting error message will have double space

emichaeli (approved):
LGTM - assuming you tested this locally

PR #307, opened 2021-11-03
SUMO-172660 Add terraform support for kinesis log sources

yuting-liu:
I've already added the new resource `sumologic_kinesis_log_source`

Already verified that the source has been created successfully:
```
resource "sumologic_kinesis_log_source" "kinesis_log" {
  name = "test-kinesis-tf"
  description = "test kinesis source with replay"
  content_type  = "KinesisLog"
  collector_id = "108404925"
  authentication {
      type = "S3BucketAuthentication"
      access_key = "XXX"
      secret_key = "XXXX"
  }
  path {
      type = "KinesisLogPath"
      bucket_name     = "test-kinesis-bucket-yuting-us-west-2"
      path_expression = "http-endpoint-failed/*"
      scan_interval   = 30000
  }
}
```

Besides, as for source update, we can update the source successfully with `NoAuthentication` placeholder:
```
resource "sumologic_kinesis_log_source" "kinesis_log" {
  name = "test-kinesis-tf"
  description = "test kinesis source with replay"
  content_type  = "KinesisLog"
  collector_id = "108404925"
  authentication {
      type = "NoAuthentication"
  }
  path {
      type = "NoPathExpression"
  }
}
```

The only issue is that there will be inconsistency after using the placeholder.
As for `terraform plan`, it will show one change needed to be apply:
```
Terraform will perform the following actions:

  # sumologic_kinesis_log_source.kinesis_log will be updated in-place
  ~ resource "sumologic_kinesis_log_source" "kinesis_log" {
        id                           = "114158598"
        name                         = "test-kinesis-tf"
        # (12 unchanged attributes hidden)


      + path {
          + scan_interval = 300000
          + type          = "NoPathExpression"
        }
        # (1 unchanged block hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.

Do you want to perform these actions?
  Terraform will perform the actions described above.
  Only 'yes' will be accepted to approve.

  Enter a value: yes

sumologic_kinesis_log_source.kinesis_log: Modifying... [id=114158598]
sumologic_kinesis_log_source.kinesis_log: Modifications complete after 2s [id=114158598]
```

However, there won't be errors.

### DIFF
--- a/sumologic/resource_sumologic_kinesis_log_source_test.go
+++ b/sumologic/resource_sumologic_kinesis_log_source_test.go
@@ -1,179 +1,179 @@
 package sumologic
 
 import (
-	"fmt"
-	"os"
-	"strconv"
-	"testing"
+    "fmt"
+    "os"
+    "strconv"
+    "testing"
 
-	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
-	"github.com/hashicorp/terraform-plugin-sdk/terraform"
+    "github.com/hashicorp/terraform-plugin-sdk/helper/resource"
+    "github.com/hashicorp/terraform-plugin-sdk/terraform"
 )
 
 func TestAccSumologicKinesisLogSource_create(t *testing.T) {
-	var kinesisLogSource KinesisLogSource
-	var collector Collector
-	cName, cDescription, cCategory := getRandomizedParams()
-	sName, sDescription, sCategory := getRandomizedParams()
-	kinesisLogResourceName := "sumologic_kinesis_log_source.kinesisLog"
-	testAwsRoleArn := os.Getenv("SUMOLOGIC_TEST_ROLE_ARN")
-	testAwsBucket := os.Getenv("SUMOLOGIC_TEST_BUCKET_NAME")
-	resource.Test(t, resource.TestCase{
-		PreCheck:     func() { testAccPreCheckWithAWS(t) },
-		Providers:    testAccProviders,
-		CheckDestroy: testAccCheckKinesisLogSourceDestroy,
-		Steps: []resource.TestStep{
-			{
-				Config: testAccSumologicKinesisLogSourceConfig(cName, cDescription, cCategory, sName, sDescription, sCategory, testAwsRoleArn, testAwsBucket),
-				Check: resource.ComposeTestCheckFunc(
-					testAccCheckKinesisLogSourceExists(kinesisLogResourceName, &kinesisLogSource),
-					testAccCheckKinesisLogSourceValues(&kinesisLogSource, sName, sDescription, sCategory),
-					testAccCheckCollectorExists("sumologic_collector.test", &collector),
-					testAccCheckCollectorValues(&collector, cName, cDescription, cCategory, "Etc/UTC", ""),
-					resource.TestCheckResourceAttrSet(kinesisLogResourceName, "id"),
-					resource.TestCheckResourceAttr(kinesisLogResourceName, "name", sName),
-					resource.TestCheckResourceAttr(kinesisLogResourceName, "description", sDescription),
-					resource.TestCheckResourceAttr(kinesisLogResourceName, "category", sCategory),
-					resource.TestCheckResourceAttr(kinesisLogResourceName, "content_type", "KinesisLog"),
-					resource.TestCheckResourceAttr(kinesisLogResourceName, "path.0.type", "KinesisLogPath"),
-				),
-			},
-		},
-	})
+    var kinesisLogSource KinesisLogSource
+    var collector Collector
+    cName, cDescription, cCategory := getRandomizedParams()
+    sName, sDescription, sCategory := getRandomizedParams()
+    kinesisLogResourceName := "sumologic_kinesis_log_source.kinesisLog"
+    testAwsRoleArn := os.Getenv("SUMOLOGIC_TEST_ROLE_ARN")
+    testAwsBucket := os.Getenv("SUMOLOGIC_TEST_BUCKET_NAME")
+    resource.Test(t, resource.TestCase{
+        PreCheck:     func() { testAccPreCheckWithAWS(t) },
+        Providers:    testAccProviders,
+        CheckDestroy: testAccCheckKinesisLogSourceDestroy,
+        Steps: []resource.TestStep{
+            {
+                Config: testAccSumologicKinesisLogSourceConfig(cName, cDescription, cCategory, sName, sDescription, sCategory, testAwsRoleArn, testAwsBucket),
+                Check: resource.ComposeTestCheckFunc(
+                    testAccCheckKinesisLogSourceExists(kinesisLogResourceName, &kinesisLogSource),
+                    testAccCheckKinesisLogSourceValues(&kinesisLogSource, sName, sDescription, sCategory),
+                    testAccCheckCollectorExists("sumologic_collector.test", &collector),
+                    testAccCheckCollectorValues(&collector, cName, cDescription, cCategory, "Etc/UTC", ""),
+                    resource.TestCheckResourceAttrSet(kinesisLogResourceName, "id"),
+                    resource.TestCheckResourceAttr(kinesisLogResourceName, "name", sName),
+                    resource.TestCheckResourceAttr(kinesisLogResourceName, "description", sDescription),
+                    resource.TestCheckResourceAttr(kinesisLogResourceName, "category", sCategory),
+                    resource.TestCheckResourceAttr(kinesisLogResourceName, "content_type", "KinesisLog"),
+                    resource.TestCheckResourceAttr(kinesisLogResourceName, "path.0.type", "KinesisLogPath"),
+                ),
+            },
+        },
+    })
 }
 func TestAccSumologicKinesisLogSource_update(t *testing.T) {
-	var kinesisLogSource KinesisLogSource
-	cName, cDescription, cCategory := getRandomizedParams()
-	sName, sDescription, sCategory := getRandomizedParams()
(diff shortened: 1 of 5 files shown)


website/docs/r/kinesis_log_source.html.markdown
```diff
@@ -0,0 +1,103 @@
...
+  }
+}
+```
+
+## Argument reference
+
+In addition to the common properties, the following arguments are supported:
+
+ - `content_type` - (Required) The content-type of the collected data. Details can be found in the [Sumologic documentation for hosted sources][1].
+ - `authentication` - (Required) Authentication details for connecting to the S3 bucket.
```
maimaisie, line 70: This should be optional. And instead of copy pasting directly from existing source page, should explain here what this S3 bucket is intended for and why someone would be interested in setting this up

website/docs/r/kinesis_log_source.html.markdown
```diff
@@ -0,0 +1,103 @@
...
+     + `type` - (Required) Must be either `S3BucketAuthentication` or `AWSRoleBasedAuthentication` or `NoAuthentication`
+     + `access_key` - (Required) Your AWS access key if using type `S3BucketAuthentication`
+     + `secret_key` - (Required) Your AWS secret key if using type `S3BucketAuthentication`
+     + `role_arn` - (Required) Your AWS role ARN if using type `AWSRoleBasedAuthentication`
+ - `path` - (Required) The location to scan for new data.
```
maimaisie, line 75: Change to optional too. Also `The location to scan for new data.` is not accurate description

website/docs/r/kinesis_log_source.html.markdown
```diff
@@ -0,0 +1,103 @@
...
+     + `type` - (Required) Must be either `KinesisLogPath` or `NoPathExpression`
+     + `bucket_name` - (Optional) The name of the bucket. This is needed if using type `S3BucketPathExpression`. 
+     + `path_expression` - (Optional) The path to the data. This is needed if using type `S3BucketPathExpression`. For Kinesis log source, it must includes `http-endpoint-failed/`.
```
maimaisie, line 78: Should not have `S3BucketPathExpression` anywhere for kinesis

website/docs/r/kinesis_log_source.html.markdown
```diff
@@ -0,0 +1,103 @@
...
+    description = "Just testing this"
+  }
+}
+```
+
+## Argument reference
+
+In addition to the common properties, the following arguments are supported:
+
+ - `content_type` - (Required) The content-type of the collected data. Details can be found in the [Sumologic documentation for hosted sources][1].
```
maimaisie, line 69: Why is this a required field if the only accepted type is Kinesis? It probably shouldn't even be exposed for the kinesis resource type

sumologic/resource_sumologic_kinesis_log_source_test.go
```diff
@@ -0,0 +1,204 @@
...
+	description = "%s"
+	category = "%s"
+	content_type  = "KinesisLog"
+	collector_id = "${sumologic_collector.test.id}"
+	authentication {
+		type = "AWSRoleBasedAuthentication"
+		role_arn = "%s"
+	  }
+	path {
+		type = "KinesisLogPath"
```
vsinghal13, line 172: nit: fix indentation

sumologic/sumologic_kinesis_log_source.go
```diff
@@ -0,0 +1,100 @@
...
+type KinesisLogResource struct {
+	ServiceType    string                   `json:"serviceType"`
+	Authentication PollingAuthentication   `json:"authentication,omitempty"`
+	Path           KinesisLogPath          `json:"path,omitempty"`
+}
+
+type KinesisLogPath struct {
+	Type              string    `json:"type"`
+    BucketName        string    `json:"bucketName,omitempty"`
+    PathExpression    string    `json:"pathExpression,omitempty"`
```
vsinghal13, line 28: nit: fix indentation. We can run `make fmt` to auto format.

website/docs/r/kinesis_log_source.html.markdown
vsinghal13: +1 this is marked as Optional in the resource.

website/docs/r/kinesis_log_source.html.markdown
vsinghal13: +1

website/docs/r/kinesis_log_source.html.markdown
yuting-liu: Will change it to `KinesisLogPath` instead

website/docs/r/kinesis_log_source.html.markdown
yuting-liu: In the kinesis metric source file, we also have this field. And here I just use the same list for properties.

maimaisie: I don't know if there is any point in exposing this field when there is only one valid value. Not sure who added this for kinesis metric source, but we shouldn't blindly follow the same mistake

maimaisie: is there multiple values possible for logs source?

yuting-liu: For polling source https://github.com/SumoLogic/terraform-provider-sumologic/blob/003ec94c82acce12d20e530b00d073e2c6079dd1/website/docs/r/polling_source.html.markdown, there will be multiple values

It seems to me that `content_type` is a field that will be use in different source types of log source resources. It doesn't seem to me as an issue considering it's a generally used property across all types of sources in sumo world.

maimaisie: But here we are dealing with kinesis log source: for this source type, are there multiple values for content type?

We need to have content type for generic source types such as polling source (which is already deprecated as indicated by the link you shared) and HTTP source, because these are both generic source types that accept different content types. However when we make a resource that is of a specific source type (such as kinesis log), if there is only one possible value for a field, then it's redundant to ask user to specify that field

maimaisie: It's unfortunate that most of the existing source types in our provider requires this content type to be specified by user even though it's not needed. But I think it's important to recognize the redundancy instead of defending it just because that's how other source types were set up. If we want to remove content type, probably we should remove it everywhere that doesn't require content type, which is a bigger effort

yuting-liu: @maimaisie I'm reading your comments. Are you fine to leave it as this in this PR with the field `contentType` to keep consistency in format for other sources as the cleanup would also require changes in other source types?

maimaisie: Yes, let's go ahead with this for now

sumologic/resource_sumologic_kinesis_log_source_test.go
maimaisie: indentation is still messed up in multiple places in this file, not sure why the conversation was marked resolved

sumologic/resource_sumologic_kinesis_log_source_test.go
```diff
@@ -0,0 +1,203 @@
...
+	category = "%s"
+	content_type  = "KinesisLog"
+	collector_id = "${sumologic_collector.test.id}"
+	authentication {
+    type = "NoAuthentication"
+  }
+	path {
+		type = "NoPathExpression"
+	}
+}
```
maimaisie, line 201: indentation

yuting-liu:
Fixed the indention manually.
Unfortunately `make fmt` doesn't seem to help the formatting 🤦‍♀️

sumologic/resource_sumologic_kinesis_log_source_test.go
```diff
@@ -0,0 +1,203 @@
...
+    content_type  = "KinesisLog"
+    collector_id = "${sumologic_collector.test.id}"
+    authentication {
+        type = "NoAuthentication"
+    }
+    path {
+        type = "NoPathExpression"
+    }
+}
+`, cName, cDescription, cCategory, sName, sDescription, sCategory)
```
maimaisie, line 202: I think this block of code is still missing one tab indentation 😓

samjsong:
> Besides, as for source update, we can update the source successfully with `NoAuthentication` placeholder

Why do we need the `NoAuthentication` placeholder? Is this also the case for our other sources that require ThirdPartyRef? Is there no way to just omit it from the update source?

sumologic/resource_sumologic_kinesis_log_source_test.go
yuting-liu: For this block, it's actually intended to format this way because it's a multiple string passed into `fmt.Sprintf()`. This is not part of the code and adding the indentation here will mess up the multiline string format itself. So we should leave it as this.

yuting-liu:
As for `NoAuthentication`, the details for why we added it can be found in description in https://github.com/Sanyaku/sumologic/pull/104210
long story in short: it helps us to update source to disable kinesis log replay.

> Is this also the case for our other sources that require ThirdPartyRef?

No.  it's something specific for Kinesis log replay

> Is there no way to just omit it from the update source?

We don't have an explicit field to decide whether kinesis log replay is enabled or disabled, but to use whether there is a third party ref block or not to determine whether to enable or disable kinesis log replay.

Because `thirdPartyRef` is an optional field, we cannot just omit it. The expected behavior of omitting one optional field in source updates is to leave it unchanged. This means we cannot disabling Kinesis log replay from enabling.

yuting-liu:
Will ping review again on this PR @maimaisie @samjsong

maimaisie (approved):

samjsong (approved):
Thank you for clarifying!

Issue #313, opened 2021-11-15
Error while getting role


SamuYoung-Factor-y:
While trying to getting an existing role name for use it for a new user this happened

```goroutine 52 [running]:
github.com/SumoLogic/terraform-provider-sumologic/sumologic.(*Client).GetRoleName(0xc000c48120, 0xc000c3e2f4, 0xa, 0xf39fe0, 0xc0007afc30, 0xc000062c01)
        github.com/SumoLogic/terraform-provider-sumologic/sumologic/data_source_sumologic_role.go:101 +0x2d2
github.com/SumoLogic/terraform-provider-sumologic/sumologic.dataSourceSumologicRoleRead(0xc000c525b0, 0x10b2de0, 0xc000c48120, 0xc000c525b0, 0x0)
        github.com/SumoLogic/terraform-provider-sumologic/sumologic/data_source_sumologic_role.go:61 +0x625
github.com/hashicorp/terraform-plugin-sdk/helper/schema.(*Resource).ReadDataApply(0xc000231ea0, 0xc00007e580, 0x10b2de0, 0xc000c48120, 0xc00012b9d8, 0x1, 0x0)
        github.com/hashicorp/terraform-plugin-sdk@v1.17.2/helper/schema/resource.go:413 +0x91
github.com/hashicorp/terraform-plugin-sdk/helper/schema.(*Provider).ReadDataApply(0xc000177000, 0xc000c5daf8, 0xc00007e580, 0xc00007e580, 0x0, 0x0)
        github.com/hashicorp/terraform-plugin-sdk@v1.17.2/helper/schema/provider.go:451 +0x8f
github.com/hashicorp/terraform-plugin-sdk/internal/helper/plugin.(*GRPCProviderServer).ReadDataSource(0xc00065e2c8, 0x12ac510, 0xc000c3d590, 0xc00089adc0, 0xc00065e2c8, 0xc000c3d590, 0xc0004c0ba0)
        github.com/hashicorp/terraform-plugin-sdk@v1.17.2/internal/helper/plugin/grpc_provider.go:1046 +0x42d
github.com/hashicorp/terraform-plugin-sdk/internal/tfplugin5._Provider_ReadDataSource_Handler(0x1079140, 0xc00065e2c8, 0x12ac510, 0xc000c3d590, 0xc000c48420, 0x0, 0x12ac510, 0xc000c3d590, 0xc000040540, 0x56)
[...]

PR #316, opened 2021-11-16
Fix provider crash for user and role data sources

emichaeli:
- Fix provider crash when role `id` / `name` not found
- Fix provider crash when user `id` / `email` not found


mccartney:
I guess this is meant to address #313, right?

troy-q-li (approved):
LGTM

Issue #317, opened 2021-11-17
`sumologic_folder` resource error: unexpected end of JSON input


devanbenz:
I am trying to use the `sumologic_folder` resource to create a folder within sumo logic to then deploy metrics dashboards. Currently, my terraform code gets caught up during the folder creation resource with the following error: 

`│ Error: unexpected end of JSON input
│ 
│   with module.sumologic.sumologic_folder.per_environment,
│   on modules/sumologic/folders.tf line 4, in resource "sumologic_folder" "per_environment":
│    4: resource "sumologic_folder" "per_environment" {`

emichaeli:
Hey @devanbenz, thanks for reporting this issue. The fix for this (which should provide a more clear error message on what went wrong) should be rolled out in the next provider release. However, I don't have enough information to know for sure whether the fix covers your case as well based on the description here.

What I believe is happening is something along those lines of:
- You've previously created that folder with Terraform, and are now using some different credentials that do not have access to this folder
- You're using the same credentials that were used to originally create the folder, but the issuing user of those credentials do not have permissions to this folder any more (for example because the user's role changed)

Do any of the above sound right? Otherwise, it would be great if you can provide more information on how you came about running to this issue. For example, is this the first time you're using the folder resource to create it? Does it only happen for a specific folder or any folder?

devanbenz:
I have tried with an existing folder (using my API credentials) and I have tried creating a folder. We have been using our primary administrator accounts credentials for the sumo api access keys as well. So I have tried both my account using my personal folder and with the administrator account using the admin recommended folder.

sumovishal:
@devanbenz would be possible for you to provide us debug logs? You can enable debug logs by prepending `TF_LOG=DEBUG`to your terraform command. The logs may help us get better understanding of the issue here.

icco:
@sumovishal could we email them? 

Deleting the folder from our (I work with @devanbenz) tf state didn't resolve the issue, but deleting the dashboards as well did, which makes me think this issue is actually unrelated to folder, and instead related to a corrupted dashboard config.

sumovishal:
@icco yeah, email is fine. Can you email it to "vishal@sumologic.com" and "emichaeli@sumologic.com"?

emichaeli:
Closing this issue as the "unexpected end of JSON input" error for both `folder` and `dashboard` resources was fixed in https://github.com/SumoLogic/terraform-provider-sumologic/pull/319

Issue #188, opened 2021-04-30
Unexpected End of JSON Input


upstart-swiss:
during the refresh step, I'm consistently getting a
Error: unexpected end of JSON input

The TF_LOG=DEBUG is mostly useless, but here's some relevant stuff from TRACE
```
2021/04/30 08:53:19 [TRACE] eval: *terraform.EvalWriteState
2021/04/30 08:53:19 [TRACE] EvalWriteState: recording 1 dependencies for module.collectors.sumologic_http_source.searchresults20200916t122321_9920700_csv_1157794899
2021/04/30 08:53:19 [TRACE] EvalWriteState: writing current state object for module.collectors.sumologic_http_source.searchresults20200916t122321_9920700_csv_1157794899
2021/04/30 08:53:19 [TRACE] [walkRefresh] Exiting eval tree: module.collectors.sumologic_http_source.searchresults20200916t122321_9920700_csv_1157794899
2021/04/30 08:53:19 [TRACE] vertex "module.collectors.sumologic_http_source.searchresults20200916t122321_9920700_csv_1157794899": visit complete
2021/04/30 08:53:19 [TRACE] vertex "module.collectors.sumologic_http_source.searchresults20200916t122321_9920700_csv_1157794899": dynamic subgraph completed successfully
2021/04/30 08:53:19 [TRACE] vertex "module.collectors.sumologic_http_source.searchresults20200916t122321_9920700_csv_1157794899": visit complete
2021/04/30 08:53:19 [TRACE] vertex "module.collectors.sumologic_http_source.searchresults20200916t122321_9920700_csv_1157794899 (expand)": dynamic subgraph completed successfully
2021/04/30 08:53:19 [TRACE] vertex "module.collectors.sumologic_http_source.searchresults20200916t122321_9920700_csv_1157794899 (expand)": visit complete
2021/04/30 08:53:20 [TRACE] eval: *terraform.EvalWriteState
2021/04/30 08:53:20 [TRACE] EvalWriteState: recording 1 dependencies for module.collectors.sumologic_http_source.<REDACTED>_api_staging
2021/04/30 08:53:20 [TRACE] EvalWriteState: writing current state object for module.collectors.sumologic_http_source.<REDACTED>_api_staging
2021/04/30 08:53:20 [TRACE] [walkRefresh] Exiting eval tree: module.collectors.sumologic_http_source.<REDACTED>_api_staging
[...]

PR #319, opened 2021-11-19
Fix unexpected end of JSON input in folder and dashboard resources

emichaeli:
Addresses https://github.com/SumoLogic/terraform-provider-sumologic/issues/188 and (probably) https://github.com/SumoLogic/terraform-provider-sumologic/issues/317.

Steps to reproduce this issue:
1. Create a folder resource using some `user1` access key/id
2. Change the access key/id to some other `user2` with minimal credentials and do `terraform plan`
3. Get `Error: unexpected end of JSON input`

The reason this happened is because we already have the folder id in the state when `user1` created it, but `user2` does not have access to it. So when `user2` does `terraform plan`, we try to read that folder id from the API and get a 404. We were not handling that 404 properly (it's when the folder in `nil`) and hence we were trying to unmarshal a `nil` json object, which results in a `unexpected end of JSON input` error.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 ## 2.12.0 (Unreleased)
 
+## 2.11.4 (November 19, 2021)
+
+BUG FIXES:
+
+* Fix unexpected end of JSON input error in folder and dashboard resources (GH-319)
+
 ## 2.11.3 (November 17, 2021)
 
 BUG FIXES:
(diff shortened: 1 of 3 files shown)


icco (approved):

icco:
Not a reviewer, but if this fixes #317, that would be great because our terraform setup is completely broken until this is resolved.

sumovishal (approved):
lgtm, let's see if it resolves 317. Regardless, we need this fix. 

Can you update the change log?

emichaeli:
I was able to reproduce this same issue with the dashboard resource as well, so included a fix for that in this PR as well.

Issue #90, opened 2020-10-05
Feature request: Local File Source support


Philipp-Navis:
Hi,

it would be great if this Terraform provider could add the support for adding Local File sources as documented here: 
https://help.sumologic.com/03Send-Data/Sources/01Sources-for-Installed-Collectors/Local-File-Source

Thanks,
Philipp

vsinghal13:
resource added in  https://github.com/SumoLogic/terraform-provider-sumologic/issues/90

PR #182, opened 2021-04-05
add sumologic_installed_collector resource

vsinghal13:
This PR adds the `sumologic_installed_collector` resource to be managed by Terraform. This can only be used to read, update and delete an already existing collector which can be imported in terraform.

- [X] Tested locally by importing an existing installed collector.
- [X] Able to assign new fields
- [X] Able to assign budget via `_budget` field
- [X] Able to delete the collector

Import statement: `terraform import sumologic_installed_collector.installed_collector 1xxxxxx`

Sample resource block:
```
resource "sumologic_installed_collector" "installed_collector" {
  name        = "vsinghal-mac-tf-new"
  category = "macos/vsinghal"
  ephemeral = true
  fields =  {
      _budget =  "Dev_20GB"
      key = "value"

    }
}
```

Sample tfstate:
```
{
  "version": 4,
  "terraform_version": "1.1.7",
  "serial": 31,
  "lineage": "ed6e03d2-f9ed-28a7-1cac-34bbdcb9160d",
  "outputs": {},
  "resources": [
    {
      "mode": "managed",
      "type": "sumologic_installed_collector",
      "name": "installed_collector",
      "provider": "provider[\"sumologic.com/dev/sumologic\"]",
      "instances": [
        {
          "schema_version": 0,
          "attributes": {
            "alive": true,
            "category": "macos/vsinghal",
            "collector_version": "19.392-5",
            "cutoff_timestamp": 0,
            "description": "",
            "ephemeral": true,
            "fields": {
              "_budget": "Dev_20GB",
              "key": "value"
            },
            "host_name": "",
            "id": "1xxxxx",
            "last_seen_alive": 1651785111049,
            "name": "vsinghal-mac-tf-new",
            "source_sync_mode": "UI",
            "target_cpu": null,
            "timezone": "Etc/UTC"
          },
          "sensitive_attributes": [],
          "private": "bnVsbA=="
        }
      ]
    }
  ]
}
```

PR #330, opened 2022-01-06
Setting up the ability to create local_file sources

amvapor:
Very Rudimentary resource provider for the resource "sumologic_local_file_source"

Implemented `path_expression` and `encoding` have not implimented `deny_list` as our use case does not yet require it

https://help.sumologic.com/03Send-Data/Sources/03Use-JSON-to-Configure-Sources/JSON-Parameters-for-Installed-Sources#Local_file_source

### DIFF
--- a/sumologic/resource_sumologic_local_file_source.go
+++ b/sumologic/resource_sumologic_local_file_source.go
@@ -71,13 +71,19 @@ func resourceSumologicLocalFileSourceUpdate(d *schema.ResourceData, meta interfa
 }
 
 func resourceToLocalFileSource(d *schema.ResourceData) LocalFileSource {
+	rawDenyList := d.Get("deny_list").(*schema.Set).List()
+	var denylist []string
+	for _, j := range rawDenyList {
+		denylist = append(denylist, j.(string))
+	}
 	source := resourceToSource(d)
 	source.Type = "LocalFile"
-	source.PathExpression = d.Get("path_expression").(string)
-	source.Encoding = d.Get("encoding").(string)
 
 	localFileSource := LocalFileSource{
-		Source: source,
+		Source:         source,
+		PathExpression: d.Get("path_expression").(string),
+		Encoding:       d.Get("encoding").(string),
+		DenyList:       denylist,
 	}
 
 	return localFileSource
@@ -105,6 +111,7 @@ func resourceSumologicLocalFileSourceRead(d *schema.ResourceData, meta interface
 	}
 	d.Set("path_expression", source.PathExpression)
 	d.Set("encoding", source.Encoding)
+	d.Set("deny_list", source.DenyList)
 
 	return nil
 }
--- a/sumologic/resource_sumologic_local_file_source.go
+++ b/sumologic/resource_sumologic_local_file_source.go
@@ -1,0 +1,110 @@
+package sumologic
+
+import (
+	"fmt"
+	"log"
+	"strconv"
+
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+)
+
+func resourceSumologicLocalFileSource() *schema.Resource {
+	localFileSource := resourceSumologicSource()
+	localFileSource.Create = resourceSumologicLocalFileSourceCreate
+	localFileSource.Read = resourceSumologicLocalFileSourceRead
+	localFileSource.Update = resourceSumologicLocalFileSourceUpdate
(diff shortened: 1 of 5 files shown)


amvapor:
May resolve https://github.com/SumoLogic/terraform-provider-sumologic/issues/90

sumologic/resource_sumologic_local_file_source.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	return resourceSumologicLocalFileSourceRead(d, meta)
+}
+
+func resourceToLocalFileSource(d *schema.ResourceData) LocalFileSource {
+	source := resourceToSource(d)
+	source.Type = "LocalFile"
+	source.PathExpression = d.Get("path_expression").(string)
+	source.Encoding = d.Get("encoding").(string)
```
maimaisie, line 77: `deny_list` seems to be missing

sumologic/resource_sumologic_local_file_source.go
```diff
@@ -0,0 +1,110 @@
...
+		d.SetId("")
+
+		return nil
+	}
+
+	if err := resourceSumologicSourceRead(d, source.Source); err != nil {
+		return fmt.Errorf("%s", err)
+	}
+	d.Set("path_expression", source.PathExpression)
+	d.Set("encoding", source.Encoding)
```
maimaisie, line 107: also missing `deny_list`

sumologic/sumologic_sources.go
```diff
@@ -30,6 +30,8 @@ type Source struct {
 	Fields                     map[string]interface{} `json:"fields,omitempty"`
 	Url                        string                 `json:"url,omitempty"`
 	ContentType                string                 `json:"contentType,omitempty"`
+	PathExpression             string                 `json:"pathExpression,omitempty"`
+	Encoding                   string                 `json:"encoding,omitempty"`
```
maimaisie, line 34: missing `deny_list`

sumologic/resource_sumologic_local_file_source_test.go
```diff
@@ -0,0 +1,259 @@
...
+	description = "%s"
+	category = "%s"
+}
+	
+resource "sumologic_local_file_source" "localFile" {
+	name = "%s"
+	description = "%s"
+	category = "%s"
+	collector_id = "${sumologic_collector.test.id}"
+}
```
maimaisie, line 241: I believe we need more fields to be added here (at least the required field `path_expression`) so the source creation will not fail

maimaisie (commented):
Thanks for submitting the PR! Made a couple comments.

sumologic/sumologic_sources.go
amvapor: I agree, I did a pass at trying to get deny_list setup for feature completion (we don't use it for our use case) But I ran into trouble with what data struct is the correct one to be used. Does anyone have a recommendation on how to implement the list? An example from another part of the codebase would also be helpful.

maimaisie: Actually I might be wrong, since path_expression, encoding and deny_list are specific to the local file source, they should not be added to the `Source` struct. @vsinghal13 can you confirm?

vsinghal13: correct, since they are specific to the local file source, these properties should only be added to the struct in `sumologic_local_file_source.go` file.

sumologic/resource_sumologic_local_file_source_test.go
vsinghal13: Also, the `sumologic_collector` is for the hosted collector. We would need to have a new resource for `installed_collector` as the `sumologic_local_file_source` is an Installed collector source.

Refer: https://github.com/SumoLogic/terraform-provider-sumologic/pull/182

zanthorn:
I would love to see this implemented.

sumologic/resource_sumologic_local_file_source_test.go
amvapor: Interesting, I may have need to leverage that installed_collector construct when it's deployed.

vsinghal13:
@amvapor please refactor this PR based on #182 to use the new `sumologic_installed_collector` resource.

vsinghal13:
closing in favor of https://github.com/SumoLogic/terraform-provider-sumologic/pull/477

PR #336, opened 2022-01-20
Fix topologyLabel to allow more than one label

sumovishal:
- Fixed topology_label field to allow more than one label
- Removed `coloring_rule` field from the doc as it is not supported. Will remove it from the code later.
- Ran `make fmt` to format `sumologic/resource_sumologic_kinesis_log_source_test.go` file (only white space changes)

### Testing
I have verified we can create more than one topology label for dashboards.
```
  topology_label_map {
    data {
      label  = "application"
      values = ["app1", "app2"]
    }
    data {
      label  = "env"
      values = ["dev", "prod"]
    }
  }
```


emichaeli (commented):
Code changes mostly LGTM, but acc tests are failing. Seems like it might be due to the ordering of elements within the topologyLabel

emichaeli (approved):
LGTM

PR #340, opened 2022-02-03
Sumo.170432.permission.api.resource

troy-q-li:
Implementation is done and verified locally. Tests are passing.
The problem is that even the user add all level of permissions when defining the resource, the order of the permission also matters.


website/docs/r/content_permission_source.html.markdown
```diff
@@ -57,13 +80,21 @@ resource "sumologic_content_permission" "content_permission_test" {
...
-- `notify_recipient` - (Required) Boolean value. Set to "true" to notify the users who had a permission update.
-- `notification_message` - (Required) The notification message sent to the users who had a permission update.
+- `content_id` - (Required) The identifier of the content item for which you want to update
+permissions.
+- `notify_recipient` - (Required) Boolean value. Set it to "true" to notify the recipients by email.
+- `notification_message` - (Optional) The notification message to send to the users.
+- `permission` - (Required) Permission block defining permission on the content. See
+[permission schema](#schema-for-permission) for details.
 
 ## Attributes reference
```
emichaeli, line 90: Delete this title as there's nothing here?

emichaeli (approved):
LGTM, please update CHANGELOG.md before merging

PR #342, opened 2022-02-07
Add import folder pass through.

zquintana:
Hello folks, 

I'm trying to update some of our orgs infrastructure to make use of Terraform. I wanted to use Terraform to manage our Sumologic instance, but I discovered this provider is missing the import setup. This is critical to migrating state from manual to Terraform. According to some people on stackoverflow this should be relatively straight forward https://stackoverflow.com/a/58376672/1072058. Could you all review this and see if it makes sense to include in your package? If it works out how soon would a new version be cut that supports this? I'd like to make use of this soon.

### DIFF
--- a/sumologic/resource_sumologic_folder.go
+++ b/sumologic/resource_sumologic_folder.go
@@ -13,9 +13,9 @@ func resourceSumologicFolder() *schema.Resource {
 		Read:   resourceSumologicFolderRead,
 		Delete: resourceSumologicFolderDelete,
 		Update: resourceSumologicFolderUpdate,
-		//		Importer: &schema.ResourceImporter{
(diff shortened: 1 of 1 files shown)


emichaeli:
Thanks @zquintana! Will take a look at this tomorrow.

zquintana:
Thanks guys. To be clear, this only for the one resource, but it's a start. I couldn't find how to test this from my fork otherwise I would have done that too. If you know how to do that I'm all ears.

emichaeli:
I created the following PR for this: https://github.com/SumoLogic/terraform-provider-sumologic/pull/345. Will be included in the next release. Thanks again @zquintana

PR #345, opened 2022-02-09
Add import functionality to folder resource

emichaeli:
Add import functionality to folder resource. Tested manually as well.

Also disabled the DEBUG logs from the retry HTTP client.


sumovishal (approved):
lgtm

PR #371, opened 2022-05-10
Adding terraform example with AWS resources

AvadheshKaria:
Description says it all, i.e. no other code change

### DIFF
--- a/website/docs/r/kinesis_metrics_source.html.markdown
+++ b/website/docs/r/kinesis_metrics_source.html.markdown
@@ -128,4 +128,282 @@ HTTP sources can be imported using the collector name and source name (`collecto
 terraform import sumologic_kinesis_metrics_source.test my-test-collector/my-test-source
 ```
 
+## Full Example (Including terraform for AWS asset creation)
+```hcl
+terraform {
+  required_providers {
+    sumologic = {
+      source = "sumologic/sumologic"
+    }
+    aws = {
+      source = "hashicorp/aws"
+    }
+  }
+}
+
+provider "sumologic" {}
+provider "aws" {}
+
+locals {
+  account_id = ""
+  aws_access_key = ""
+  aws_secret_key = ""
+
+  description = "update your terraform description here"
+  identifier = "SumologicMetricStream"
+
+  region = "us-west-2"
+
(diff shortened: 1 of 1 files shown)


website/docs/r/kinesis_metrics_source.html.markdown
```diff
@@ -128,4 +128,282 @@ HTTP sources can be imported using the collector name and source name (`collecto
...
+  role_arn      = aws_iam_role.metric_stream_to_firehose.arn
+  firehose_arn  = aws_kinesis_firehose_delivery_stream.kinesis_stream.arn
+  output_format = "opentelemetry0.7"
+
+//  include_filter {
+//    namespace = "AWS/ApplicationELB"
+//  }
+//  include_filter {
+//    namespace = "AWS/DynamoDB"
+//  }
```
vsinghal13, line 207: does this need to be uncommented?

website/docs/r/kinesis_metrics_source.html.markdown
```diff
@@ -128,4 +128,282 @@ HTTP sources can be imported using the collector name and source name (`collecto
...
+      for_each = local.tagfilters
+      content {
+        type      = tag_filters.value.type
+        namespace = tag_filters.value.namespace
+        tags      = tag_filters.value.tags
+      }
+    }
+  }
+}
+
```
vsinghal13, line 193: should we remove the sumologic resources as they are already mentioned above in the Example usage section?

website/docs/r/kinesis_metrics_source.html.markdown
AvadheshKaria: I added comment to explain the lines, i.e. I think adding commented example is good in this case

website/docs/r/kinesis_metrics_source.html.markdown
AvadheshKaria: I am thinking if someone just does copy and paste of the full section it should work.
I was working with customer, and I feel copy paste should work for one example end-to-end.

vsinghal13 (approved):

kagarwal06 (approved):

PR #374, opened 2022-05-12
Add data source for Folders

### LINKED ISSUE #48: Support Data source for Folders apart from Personal_folder
Currently I see 'sumologic_personal_folder' data source can get you the ID for personal folder and I can pass that as parent id while creating any scheduled search. But it will be useful if we could find ID for any Folder (not just personal folder).
Are you planning to support Data source say 'sumologic_folder' as a generic one to get ID for any folder we pass ?
Otherwise, we have no option but to use Sumologic's REST API to get the ID for the folder first and then pass it while creating scheduled search, which is really not ideal.

sumovishal:
Adding a new data source for folder that allows user to refer to existing folders in the org. Tested against us2 test org.

Fixes #48


dagould (approved):
lgtm

PR #382, opened 2022-05-20
Updating minor version for release 2.16.0

arpitj-sumo:


CHANGELOG.md
```diff
@@ -1,7 +1,7 @@
-## 2.16.0 (Unreleased)
+## 2.16.1 (Unreleased)
```
dagould, line 1: I think we usually mark the Unreleased placeholder as a "minor" rev, i.e., it'd be 2.17.0 in this case.
(that being the most likely next version; it can be changed to 2.16.1 or something else if needed.)

arpitj-sumo: Fixed.

dagould (approved):
lgtm

PR #392, opened 2022-06-13
Update sumologic_sources.go

### LINKED ISSUE #259: Sources idempotency bug and/or not clear on how to use "Use Collector Default" timezone
I'd like to define a HTTP source which has "Use Collector Default" timezone using Terraform. In the UI you do that by setting this:
image

## Steps to reproduce
1. Run `terraform apply` with:
```
resource "sumologic_collector" "collector" {
  name        = "2021-08-mccartney-testing"
  description = "Just testing collectors' Terraform"
  timezone = "Europe/Warsaw"
}


resource "sumologic_http_source" "important-logs" {
  name                = "mccartney-logs"
  description         = ""
  category            = "something"
  collector_id        = sumologic_collector.collector.id
}
```
2. Observe the source created. It default the timezone to:

> (UTC) Etc/UTC

3. Manually edit the source to use "Use Collector Default" in the UI
4. Subsequent `terraform apply` insists on setting the timezone to UTC:
```
Terraform will perform the following actions:

  # sumologic_http_source.important-logs will be updated in-place
  ~ resource "sumologic_http_source" "important-logs" {
        id                           = "180026286"
        name                         = "mccartney-logs"
      + timezone                     = "Etc/UTC"
        # (10 unchanged attributes hidden)
    }

Plan: 0 to add, 1 to change, 0 to destroy.
```

## Context

I think the Terraform provider conflates default (UTC) with another default ("Use Collector Default")

kevin-sumo:
Fixes #259. Timezone is an optional field and default should be NIL. When set to NIL the timezone will be taken from the Collector, which is the standard default when a timezone is not set on a source.

### DIFF
--- a/sumologic/sumologic_sources.go
+++ b/sumologic/sumologic_sources.go
@@ -74,7 +74,7 @@ func resourceSumologicSource() *schema.Resource {
 			"timezone": {
 				Type:     schema.TypeString,
 				Optional: true,
-				Default:  "Etc/UTC",
+				Default:  nil,
 			},
 			"automatic_date_parsing": {
 				Type:     schema.TypeBool,
--- a/sumologic/sumologic_sources.go
+++ b/sumologic/sumologic_sources.go
@@ -74,7 +74,7 @@ func resourceSumologicSource() *schema.Resource {
 			"timezone": {
 				Type:     schema.TypeString,
 				Optional: true,
-				Default:  nil,
+				Default:  "",
 			},
 			"automatic_date_parsing": {
 				Type:     schema.TypeBool,


sumologic/sumologic_sources.go
```diff
@@ -74,7 +74,7 @@ func resourceSumologicSource() *schema.Resource {
 			"timezone": {
 				Type:     schema.TypeString,
 				Optional: true,
-				Default:  "Etc/UTC",
+				Default:  nil,
```
vsinghal13, line 77: this needs to be set to empty string `""`

vsinghal13 (changes requested):

sumologic/sumologic_sources.go
kevin-sumo: Updated nil to ""

vsinghal13 (approved):

vsinghal13:
@sumovishal can we get this merged?

PR #401, opened 2022-07-21
add use_versioned_api parameter for s3 source

vsinghal13:
We had recently exposed a new parameter for Sumologic S3 source to specify whether to use the AWS Versioned API's or not while fetching the data from the S3 bucket.

Adding support for the same in this PR. The default value of this parameter is `true` and it is optional.

### DIFF
--- a/sumologic/resource_sumologic_generic_polling_source.go
+++ b/sumologic/resource_sumologic_generic_polling_source.go
@@ -125,6 +125,11 @@ func resourceSumologicGenericPollingSource() *schema.Resource {
 					Type:     schema.TypeString,
 					Optional: true,
 				},
+				"use_versioned_api": {
+					Type:     schema.TypeBool,
+					Optional: true,
+					Default:  false,
+				},
 				"path_expression": {
 					Type:     schema.TypeString,
 					Optional: true,
@@ -336,6 +341,7 @@ func getPollingThirdPartyPathAttributes(pollingResource []PollingResource) []map
 			"limit_to_regions":              t.Path.LimitToRegions,
 			"limit_to_namespaces":           t.Path.LimitToNamespaces,
 			"limit_to_services":             t.Path.LimitToServices,
+			"use_versioned_api":             t.Path.UseVersionedApi,
 			"custom_services":               flattenCustomServices(t.Path.CustomServices),
 			"tag_filters":                   flattenPollingTagFilters(t.Path.TagFilters),
 			"sns_topic_or_subscription_arn": flattenPollingSnsTopicOrSubscriptionArn(t.Path.SnsTopicOrSubscriptionArn),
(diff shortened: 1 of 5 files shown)


sumologic/resource_sumologic_generic_polling_source.go
```diff
@@ -125,6 +125,11 @@ func resourceSumologicGenericPollingSource() *schema.Resource {
 					Type:     schema.TypeString,
 					Optional: true,
 				},
+				"use_versioned_api": {
+					Type:     schema.TypeBool,
+					Optional: true,
+					Default:  false,
```
maimaisie, line 131: In UI we have true as default value though, which one is correct?

vsinghal13: the default is `true`. Fixed.

AvadheshKaria:
Should we add this flag as part of tests

sumologic/resource_sumologic_generic_polling_source.go
```diff
@@ -590,6 +597,7 @@ func getPollingPathSettings(d *schema.ResourceData) (PollingPath, error) {
 					LimitToNamespaces = append(LimitToNamespaces, v.(string))
 				}
 			}
+			pathSettings.UseVersionedApi = false
```
maimaisie, line 600: does this need to be set to false for all path except `S3BucketPathExpression`? just wondering why other path properties like `TagFilters` that also only applies to subset of path types are not set to a bogus value in path types that don't apply

vsinghal13: other properties don't have any default value assigned to them as we have for `use_versioned_api` parameter, that is why I have explicitly set it to false for other paths.

maimaisie (approved):

sumologic/resource_sumologic_generic_polling_source.go
maimaisie: why is that? tag filters is also an optional field on cloudwatch source for example

vsinghal13: even though `tag_filters` is optional, it does not have a default value assigned to it. We can also try removing the default value for `use_versioned_api` and only set it if required.

maimaisie: I think that makes more sense because we don't want to send `false` to other source types that don't recognize this property. just need to make sure if user doesn't specify it for S3 source it will still default to `false` on the backend

maimaisie: also, the default should be true

website/docs/r/s3_source.html.markdown
```diff
@@ -59,6 +59,7 @@ In addition to the [Common Source Properties](https://registry.terraform.io/prov
      + `bucket_name` - (Required) The name of the bucket. 
      + `path_expression` - (Required) The path to the data.
      + `sns_topic_or_subscription_arn` - (Computed) This is a computed field for SNS topic/subscription ARN.
+     + `use_versioned_api` - (Optional) Whether to Use AWS versioned APIs. Default is set to `true`. If you're collecting from a Cisco Umbrella bucket this must be set to `false`.
```
yuting-liu, line 62: typo:
```
Whether to use AWS versioned APIs.
```

yuting-liu: Sorry for reviewing late 🤦

PR #403, opened 2022-07-25
SUMO-196857: adding alert grouping support for monitors

dstephens5:
This adds support for configuring notification group fields via terraform. 

This is tested in units tests and I have created a test monitor with notificationGroupFields on stag


website/docs/r/monitor.html.markdown
```diff
@@ -320,6 +322,7 @@ The following arguments are supported:
 - `group_notifications` - (Optional) Whether or not to group notifications for individual items that meet the trigger condition. Defaults to true.
 - `playbook` - (Optional - Beta) Notes such as links and instruction to help you resolve alerts triggered by this monitor. {{Markdown}} supported. It will be enabled only if available for your organization. Please contact your Sumo Logic account team to learn more.
 - `alert_name` - (Optional) The display name when creating alerts. Monitor name will be used if `alert_name` is not provided. All template variables can be used in `alert_name` except `{{AlertName}}` and `{{ResultsJson}}`.
+- `notification_group_fields` - (Optional - Beta) Set of fields used to group data to create independent notifications and alerts.
```
tarunk2, line 325: `The set of fields to be used to group alerts and notifications for a monitor. The value of this field will be considered only when 'groupNotifications' is true` - we can be consistent with api docs.

tarunk2 (approved):

website/docs/r/monitor.html.markdown
dstephens5: done

gchairuangsang (approved):
lgtm

PR #410, opened 2022-08-03
SUMO-189150: removing validations from tf for SLI window aggregation

arpitjain305:
https://jira.kumoroku.com/jira/browse/SUMO-189150

Already removed the check from Swagger API as it is being done on backend. Can remove the explicit validation from TF provider too.


sumologic/resource_sumologic_slo.go
```diff
@@ -8,15 +8,13 @@ import (
 	"regexp"
 )
 
-const sloAggregationRegexString = `^(Avg|Min|Max|Sum|(p[5-9][0-9])(\.\d{1,3})?$)$` // TODO update it to allow lower pct values
-const sloAggregationWindowRegexString = `^[0-9]{1,2}(m|h)$`                        // TODO make it exact of min 1m and max 1h
+const sloAggregationWindowRegexString = `^[0-9]{1,2}(m|h)$` // TODO make it exact of min 1m and max 1h
```
apoorv-garg, line 11: why are we keeping this validation ?

arpitjain305: hmm, we can remove this. It is being done in backend already, right?

apoorv-garg: Yep

apoorv-garg (approved):

PR #415, opened 2022-08-10
SUMO-199451 update monitor doc - disallow playbook temp var in alertname

gchairuangsang:


tarunk2 (approved):
Let's merge this after `{{Playbook}}` variable support is released.

PR #424, opened 2022-08-25
fix slo compliance period validation

apoorv-garg:


sumologic/resource_sumologic_slo.go
```diff
@@ -684,3 +682,36 @@ func verifySLOObject(slo SLOLibrarySLO) error {
...
+	var validPeriods []string
+
+	calendarPeriods := []string{"Week", "Month", "Quarter"}
+	for _, calendarPeriod := range calendarPeriods {
+		validPeriods = append(validPeriods, calendarPeriod)
+	}
+
+	minRollingDays := 1
+	maxRollingDays := 90
+	for days := minRollingDays; days <= maxRollingDays; days++ {
```
agaurav, line 703: it will be cleaner to just use the regex `^([1-9]|[1-8][0-9]|90)d$` instead.
verified here : https://regex101.com/r/43nFBe/1

agaurav (approved):
LGTM.

sumologic/resource_sumologic_slo.go
apoorv-garg: I was thinking of doing that, but then I feel it is easier to modify this, and is more straightforward to read (I hardly use regexes so it takes me time to grok them). For example: if we decide to increase this limit to 180 days, we will just need to do `maxRollingDays := 180`

Issue #428, opened 2022-09-01
Regression in 2.18.1 on multi-line monitor descriptions


mccartney:
Related to #420.

The following code works fine in `2.18.0` (applies successfully, monitor is created), but fails in `2.18.1`:
```
resource "sumologic_monitor" "tf_logs_monitor_1" {
  name         = "Terraform Logs Monitor"
  description  = trimspace(<<DESCRIPTION
Here
comes
a
long
description
DESCRIPTION
  )
  type         = "MonitorsLibraryMonitor"
  is_disabled  = false
  content_type = "Monitor"
  monitor_type = "Logs"
  evaluation_delay = "5m"

  queries {
    row_id = "A"
    query  = "_sourceCategory=event-action info"
  }

  trigger_conditions {
    logs_static_condition {
      critical {
        time_range = "15m"
        alert {
          threshold      = 40.0
          threshold_type = "GreaterThan"
        }
        resolution {
          threshold      = 40.0
          threshold_type = "LessThanOrEqual"
        }
      }
    }
  }
  playbook = "{{Name}} should be fixed in 24 hours when {{TriggerType}} is triggered."
  alert_name = "Alert from {{Name}}"
  notification_group_fields = ["_sourceHost"]
}
```

upstart-swiss:
Bug in same area - doing a blank description also fails this:
```
description = ""
```

PR #429, opened 2022-09-01
SUMO-200871 fixing the regex for not allowing leading and trailing whitespaces in monitor name and desc

tarunk2:
Terraform [regex](https://www.terraform.io/language/functions/regex) doesn't match newline with `.`. It needs a special matching flag for it. Adding it to validation regexes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 2.18.2 (September 1, 2022)
+
+BUG FIXES:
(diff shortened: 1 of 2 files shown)


mccartney (approved):

mccartney:
That should close #428, right? @tarunk2

tarunk2:
There is one more edge case. The current version doesn't allow empty or one-character descriptions. Users have to drop the description field entirely instead of specifying an empty description. We can keep it open until we fix this scenario as well.

PR #442, opened 2022-10-03
Add suppression diff rule for time formats

MZochowskiSUMO:
This PR adds suppression rule for "normal" CRUD TF resources. The case for `sumologic_content` and its JSONs was filed as a separate issue: https://github.com/SumoLogic/terraform-provider-sumologic/issues/444

### DIFF
--- a/sumologic/util.go
+++ b/sumologic/util.go
@@ -3,6 +3,7 @@ package sumologic
 import (
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+	"strconv"
 )
 
 var (
@@ -293,4 +294,52 @@ func GetTimeRangeBoundary(tfRangeBoundary map[string]interface{}) interface{} {
 	}
 
 	return nil
+}
+
+func SuppressEquivalentTimeDiff(k, oldValue, newValue string, d *schema.ResourceData) bool {
+	return getCanonicalTime(newValue) == getCanonicalTime(oldValue)
+}
+
+func getCanonicalTime(time string) string {
+	var negative bool = false
+	var absTime string = time
+	if time[0] == '-' {
+		negative = true
+		absTime = time[1:]
+	}
+
+	seconds, err := strconv.Atoi(absTime[:len(absTime)-1])
+	if err != nil {
+		return time
+	}
+
+	switch absTime[len(absTime)-1:] {
+	case "m":
+		seconds *= 60
+	case "h":
+		seconds *= 3600
+	case "d":
+		seconds *= 86400
+	case "w":
+		seconds *= 604800
+	}
+
+	var newTime string
+
+	switch {
+	case seconds%604800 == 0:
+		newTime = strconv.Itoa(seconds/604800) + "w"
+	case seconds%86400 == 0:
+		newTime = strconv.Itoa(seconds/86400) + "d"
+	case seconds%3600 == 0:
+		newTime = strconv.Itoa(seconds/3600) + "h"
+	case seconds%60 == 0:
+		newTime = strconv.Itoa(seconds/60) + "m"
+	}
+
+	if negative {
+		return "-" + newTime
+	}
+
+	return newTime
 }
--- a/sumologic/util.go
+++ b/sumologic/util.go
@@ -303,6 +303,22 @@ func GetTimeRangeBoundary(tfRangeBoundary map[string]interface{}) interface{} {
 	return nil
 }
 
+/*
+This function returns a function (in 2 variants) determining whether two stringe representation
+of a time value can be considered equivalent. One variant compares only absolute values (ignores '-' sign)
+and the other compare relative values. For details see util_test.go::TestSuppressTimeDiff.
+
+Some examples (we accept time units for seconds, minutes, hours, days and weeks):
+-1h = -60m
+1h20m = 80m
+60m60m60m1h = 3h30m30m
+-60m = 1h (only if we compare absolute values, so with isRelative = false)
+1w = 604800s
+1h != 61m
+2m != 119s
+-1h != 1h (only if we compare relative values, so with isRelative = true)
+1d = 22h60m3600s
+*/
 func SuppressEquivalentTimeDiff(isRelative bool) func(k, oldValue, newValue string, d *schema.ResourceData) bool {
 	return func(k, oldValue, newValue string, d *schema.ResourceData) bool {
 		var handleError = func(err error) bool {
--- a/sumologic/util.go
+++ b/sumologic/util.go
@@ -176,7 +176,7 @@ func GetTimeRangeBoundarySchema() map[string]*schema.Schema {
 					"relative_time": {
 						Type:             schema.TypeString,
 						Required:         true,
-						DiffSuppressFunc: SuppressEquivalentTimeDiff,
+						DiffSuppressFunc: SuppressEquivalentTimeDiff(true),
 					},
 				},
 			},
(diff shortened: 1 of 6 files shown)


sumologic/util.go
```diff
@@ -294,3 +301,96 @@ func GetTimeRangeBoundary(tfRangeBoundary map[string]interface{}) interface{} {
...
+		var handleError = func(err error) bool {
+			log.Printf("[WARN] Time value could not be converted to seconds. Details: %s", err.Error())
+			return false
+		}
+
+		if oldValue == "" || newValue == "" {
+			return false
+		}
+
+		if oldValInSeconds, oldErr := getTimeInSeconds(oldValue); oldErr != nil {
```
MZochowskiSUMO, line 316: Is this the correct "go" way of handling errors?

sumologic/util.go
```diff
@@ -294,3 +301,96 @@ func GetTimeRangeBoundary(tfRangeBoundary map[string]interface{}) interface{} {
...
+	case "m":
+		seconds *= 60
+	case "h":
+		seconds *= 3600
+	case "d":
+		seconds *= 86400
+	case "w":
+		seconds *= 604800
+	default:
+		return 0, fmt.Errorf("Only [smhdw] time units are supported, but got %s", timeValue[len(timeValue)-1:])
```
MZochowskiSUMO, line 384: I had to return sth, so I decided to return 0. Is there a better way to do it?

sumologic/util_test.go
```diff
@@ -0,0 +1,202 @@
+package sumologic
+
+import (
+	"testing"
+)
+
+// Based on https://github.com/hashicorp/terraform-provider-aws/blob/232c3ed9c9d18aab6f6b70672d7c46d44c75a52e/internal/verify/diff_test.go#L62.
+func TestSuppressTimeDiff(t *testing.T) {
```
MZochowskiSUMO, line 8: I added a unit test here.
I am not sure we can create an acceptance test for the case where we want to assert that the plan was empty. If you know how to do it, please let me know, I will write such acceptance test.

sumologic/util.go
```diff
@@ -294,3 +301,96 @@ func GetTimeRangeBoundary(tfRangeBoundary map[string]interface{}) interface{} {
 
 	return nil
 }
+
+func SuppressEquivalentTimeDiff(isRelative bool) func(k, oldValue, newValue string, d *schema.ResourceData) bool {
```
sumovishal, line 305: A function header listing out examples of time differences we are suppressing would be helpful.

sumologic/util_test.go
```diff
@@ -0,0 +1,202 @@
+package sumologic
```
sumovishal, line 1: Unit test is great. Let's update some acceptance tests in `resource_sumologic_dashboard_test.go` to make sure we have coverage in acceptance tests too.

sumovishal (commented):
Made first pass, left some comments.

sumologic/util_test.go
MZochowskiSUMO: I thought about it, but I didn't find a way to test that the plan is empty. Do you know how to write such a test?

MZochowskiSUMO: I can add a positive/negative variant (whichever lacks) there, but I don't see how it tests this PR, which only changes the emptiness of a plan.

sumologic/util.go
sumovishal: yeah, a comment above the function.

sumologic/util_test.go
sumovishal: Yes to what Greg said. 

>I can add a positive/negative variant (whichever lacks) there, but I don't see how it tests this PR

With this PR, there should no diff in the acceptance test if we use one of the variants. Like `-24h` or `-1d` should not show any diff in acceptance test.

MZochowskiSUMO: Hi, I agree that there is a value in such tests. However, as I mentioned I didn't find a way to assert that a test is empty (only that it is not empty). Do you know how to do it/whether it is possible?

sumovishal: >I didn't find a way to assert that a test is empty (only that it is not empty). 

You don't need to assert for anything. Having one of the time range variant like `-24h` in the acceptance test will ensure new code is tested. The backend will return `-1d` in such case, and the custom diff function will make sure there is no diff between state and the tf file.

MZochowskiSUMO: Thanks, I didn't know that the acc test already checks whether the plan was empty (maybe that's why I didn't find any flag to enforce such check :) ). I modified tests for dashboards, but I will also do it for monitors, as my change is applicable also there.

sumologic/util.go
```diff
@@ -294,3 +301,96 @@ func GetTimeRangeBoundary(tfRangeBoundary map[string]interface{}) interface{} {
...
+	return func(k, oldValue, newValue string, d *schema.ResourceData) bool {
+		var handleError = func(err error) bool {
+			log.Printf("[WARN] Time value could not be converted to seconds. Details: %s", err.Error())
```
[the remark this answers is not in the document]
sumovishal: >When I tested it I didn't get any such logs (here and in the other parts of our provider).

You need to prefix terraform commands with `TF_LOG=DEBUG` to see these logs.

sumologic/util.go
sumovishal: yeah, this looks right to me.

sumovishal (approved):
LGTM, please wait for a review from the Alerting team. cc @rmamilla-sumo @tarunk2

sumologic/util.go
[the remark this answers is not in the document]
MZochowskiSUMO: thanks!

sumologic/util.go
```diff
@@ -296,3 +302,112 @@ func GetTimeRangeBoundary(tfRangeBoundary map[string]interface{}) interface{} {
...
+/*
+This function returns a function (in 2 variants) determining whether two stringe representation
+of a time value can be considered equivalent. One variant compares only absolute values (ignores '-' sign)
+and the other compare relative values. For details see util_test.go::TestSuppressTimeDiff.
+
+Some examples (we accept time units for seconds, minutes, hours, days and weeks):
+-1h = -60m
+1h20m = 80m
+60m60m60m1h = 3h30m30m
+-60m = 1h (only if we compare absolute values, so with isRelative = false)
```
rmamilla-sumo, line 315: @MZochowskiSUMO , Is user allowed to set isRelative field ? why are we allowing it if there is no difference if user enters -ve / +ve value

MZochowskiSUMO: @rmamilla-sumo No it is not available to users. isRelative is only a flag we use in code (to suppress diffs) to specify if we treat -ve / +ve values as the same value. In most places there are equivalent, but here afaik they are not: https://github.com/SumoLogic/terraform-provider-sumologic/pull/442/files#diff-4bc5e99d3cf4cbda61368cb6fdbafe1bc97884202cca74d578e12b053064cb25R176, https://github.com/SumoLogic/terraform-provider-sumologic/pull/442/files#diff-b749eb8ae5017ef0ed2545e094d7e38321f015a91b335265140abcf75fb61365R503 (relativeTimeRangeSchema). Can you confirm that?

MZochowskiSUMO: @rmamilla-sumo We are suppressing time diff values treating -ve/+ve as equal everywhere except the 2 places above. isRelative flag only changes the semantics of diff suppression, it does not enable/disable it.

rmamilla-sumo (approved):
LGTM

tarunk2 (approved):

sumologic/util.go
MZochowskiSUMO: I changed suppression to absolute (isRelative = false) in SLO monitors' trigger_conditions.

PR #443, opened 2022-10-03
Make validation for time values more restrictive

MZochowskiSUMO:
@sumovishal Is this a potentially breaking change for the existing configurations? I am thinking about a scenario when someone had a "wrong" configuration with respect to the new validation rule? Is there some process to safely introduce new validation for existing fields?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,10 @@
-ENHANCEMENTS:
-* Add better validation and documentation for some time attributes (GH-443)
-
 ## 2.19.1 (Unreleased)
 FEATURES:
 * Add new optional `resolution_window` field to resource/sumologic_monitor (GH-418)
 BUG FIXES:
 * cse rules hard failing if passing tags with empty strings. (GH-445)
+ENHANCEMENTS:
+* Add better validation and documentation for some time attributes (GH-443)
 
 ## 2.19.0 (September 20, 2022)
 FEATURES:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+BUG FIXES:
+* Add better validation and documentation for some time attributes.
+
 ## 2.19.1 (Unreleased)
 FEATURES:
 * Add new optional `resolution_window` field to resource/sumologic_monitor (GH-418)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## 2.19.1 (Unreleased)
 FEATURES:
 * Add new optional `resolution_window` field to resource/sumologic_monitor (GH-418)
+BUG FIXES:
+* cse rules hard failing if passing tags with empty strings. (GH-445)
 
 ## 2.19.0 (September 20, 2022)
 FEATURES:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## 2.19.2 (Unreleased)
+ENHANCEMENTS:
+* Add better validation and documentation for some time attributes (GH-443)
 
 ## 2.19.1 (October 6, 2022)
 FEATURES:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
-BUG FIXES:
-* Add better validation and documentation for some time attributes.
+ENHANCEMENTS:
+* Add better validation and documentation for some time attributes (GH-443)
 
 ## 2.19.1 (Unreleased)
 FEATURES:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,12 @@
-## 2.19.1 (Unreleased)
+## 2.19.2 (Unreleased)
+
+## 2.19.1 (October 6, 2022)
 FEATURES:
 * Add new optional `resolution_window` field to resource/sumologic_monitor (GH-418)
+
 BUG FIXES:
-* cse rules hard failing if passing tags with empty strings. (GH-445)
-ENHANCEMENTS:
-* Add better validation and documentation for some time attributes (GH-443)
+* CSE rules hard failing if passing tags with empty strings. (GH-445)
+* Return error when unable to read collectors. (GH-446)
 
 ## 2.19.0 (September 20, 2022)
 FEATURES:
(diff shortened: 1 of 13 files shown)


mccartney (approved):

MZochowskiSUMO:
Added new commit that prohibits values like 30min, 4hours, etc. on multiple 'time' fields. Previously we expected only numbers and first letters of second, minute, etc. It could have led to valid values like "30mThisPartIsNotValidated".

mccartney (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+BUG FIXES:
```
MZochowskiSUMO, line 1: @sumovishal I didn't know where to put it. It can also be classified as enhancement (?).

sumovishal:
>@sumovishal Is this a potentially breaking change for the existing configurations?

CHANGELOG.md
sumovishal: Let's put them under "ENHANCEMENTS".

MZochowskiSUMO:
> > @sumovishal Is this a potentially breaking change for the existing configurations?
> 
> Were these validations present in the API? I assume so, in that case, this should not be a breaking change as user would fail to create the monitor with invalid time range.

We think so.

MZochowskiSUMO:
Please ignore that new files/commits were added "in the past". I combined rebase and merge when updating branch to the current version of master :(
My new changes are only the acceptance tests and CHANGE LOG ones.

sumologic/resource_sumologic_monitors_library_monitor_test.go
```diff
@@ -106,6 +106,43 @@ func TestAccSumologicMonitorsLibraryMonitor_schemaValidations(t *testing.T) {
 	})
 }
 
+func TestAccSumologicMonitorsLibraryMonitor_timeFieldsSchemaValidation(t *testing.T) {
```
MZochowskiSUMO, line 109: @sumovishal Should I make acc test for all the fields I added validation for? It would be nice to have, but in my understanding (very basic) acc tests are costly.

sumovishal: Yeah, I would not recommend an acceptance test for the validation. If you want to, I would suggest adding a unit test.

sumovishal:
> > > @sumovishal Is this a potentially breaking change for the existing configurations?
> > 
> > 
> > Were these validations present in the API? I assume so, in that case, this should not be a breaking change as user would fail to create the monitor with invalid time range.
> 
> We think so. I will proceed with the change.

Can you confirm with the respective teams? For monitor, @tarunk2 should be able to help you. For CSE resource, ask @pmontiel-sumo to confirm the validations are present on the API side.

sumologic/resource_sumologic_monitors_library_monitor_test.go
MZochowskiSUMO: Ok, I will remove it.

MZochowskiSUMO:
> 

I didn't add validation for CSE, it was my mistake with rebase I mentioned earlier. I added it to dashboards (already reached out to the team) and to monitors.

sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -107,7 +107,7 @@ func resourceSumologicMonitorsLibraryMonitor() *schema.Resource {
 						"time_range": {
 							Type:         schema.TypeString,
 							Optional:     true,
-							ValidateFunc: validation.StringMatch(regexp.MustCompile(`-?(\d)+[smhd]`), "Time range must be in the format '-?\\d+[smhd]'. Examples: -15m, 1d, etc."),
+							ValidateFunc: validation.StringMatch(regexp.MustCompile(`^-?(\d)+[smhd]$`), "Time range must be in the format '-?\\d+[smhd]'. Examples: -15m, 1d, etc."),
```
rmamilla-sumo, line 110: we don't want to force user to use -ve values ATM. 

Current logic basically takes -ve/+ve values but when the response comes from api, we are converting the time range to a positive value this will lead to some issues where tf state would not match and tf will keep applying.

website/docs/r/monitor.html.markdown
```diff
@@ -400,26 +406,26 @@ Here is a summary of arguments for each condition type (fields which are not mar
 #### logs_static_condition
   - `field`
   - `critical`
-    - `time_range` (Required)
+    - `time_range` (Required) : Accepted format: Optional `-` sign followed by `<number>` followed by a `<time_unit>` character: `s` for seconds, `m` for minutes, `h` for hours, `d` for days. Examples: `30m`, `-12h`.
```
rmamilla-sumo, line 409: currently i think users are leveraging +ve values not -ve. see my comment above

rmamilla-sumo (changes requested):
enforcing -ve on time_range will not work atm. plz revert them

sumologic/resource_sumologic_monitors_library_monitor.go
mccartney: > we don't want to force user to use -ve values ATM.

How is the change forcing the `-` sign? I fail to see it.

MZochowskiSUMO: I agree that having choice here between +ve/-ve will lead to TF problems (state won't change after apply), but it is solved in this PR: https://github.com/SumoLogic/terraform-provider-sumologic/pull/442. Additionally, as Greg said, new validation does not force sign, it only forces that the old regex matches whole value, not a substring as before.

website/docs/r/monitor.html.markdown
MZochowskiSUMO: I commented above that this is not the case. Here I added a description for the field format, with optional `-` sign.

sumologic/resource_sumologic_monitors_library_monitor.go
rmamilla-sumo: Thanks @MZochowskiSUMO for the explanation.
So as per #442 , we are basically comparing abs values.
regex: `^-?(\d)+[smhd]$` has ? after - which i didn't notice, makes sense that we are not forcing

MZochowskiSUMO: I will mark it then as resolved.

rmamilla-sumo (approved):

PR #446, opened 2022-10-04
return error in case of unable to read collector

vsinghal13:
recently, we found a bug, where if we are not able to read a collector for eg. due to 5xx, TF would consider that the collector does not exist and clean up the collector resource. That also cleans up the attached sources since the sources have a parent-child entity relationship.


mccartney (approved):

sumovishal:
L63 needs the same fix.

PR #458, opened 2022-11-09
Sumo 203323 terraform min data points field

ndhar-sumo:
A new optional field for metrics static condition - min data points field

Testing:
Created monitor on Stag using Tf definition


ndhar-sumo:
The acceptance test is failing - after debugging i think i might need to change the schema or the test. The monitor gets created correctly on Stag.

sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -120,6 +120,11 @@ func resourceSumologicMonitorsLibraryMonitor() *schema.Resource {
 							Optional:     true,
 							ValidateFunc: validation.StringInSlice([]string{"AtLeastOnce", "Always", "ResultCount", "MissingData"}, false),
 						},
+						"min_data_points": {
+							Type:         schema.TypeInt,
+							Optional:     true,
+							ValidateFunc: validation.IntAtLeast(1),
```
dstephens5, line 126: Do we want to validate this being less than 100 as well?

dstephens5 (approved):
one comment, otherwise LGTM since this has been tested and is working on stag

sumologic/resource_sumologic_monitors_library_monitor.go
ndhar-sumo: Yes!

PR #472, opened 2023-01-20
TRIDENT-16345: Support CSE FirstSeenRule

f49009bc-aa98-42c5-b7d3-5b2311812db4:
Issues: TRIDENT-16345

### DIFF
--- a/sumologic/resource_sumologic_cse_first_seen_rule.go
+++ b/sumologic/resource_sumologic_cse_first_seen_rule.go
@@ -1,0 +1,213 @@
+package sumologic
+
+import (
+	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/helper/validation"
+	"log"
+)
+
+func resourceSumologicCSEFirstSeenRule() *schema.Resource {
+	return &schema.Resource{
+		Create: resourceSumologicCSEFirstSeenRuleCreate,
+		Read:   resourceSumologicCSEFirstSeenRuleRead,
+		Delete: resourceSumologicCSEFirstSeenRuleDelete,
+		Update: resourceSumologicCSEFirstSeenRuleUpdate,
+		Importer: &schema.ResourceImporter{
+			State: schema.ImportStatePassthrough,
+		},
+
+		Schema: map[string]*schema.Schema{
+			"baseline_type": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"baseline_window_size": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"category": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"description_expression": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"enabled": {
+				Type:     schema.TypeBool,
+				Required: true,
+			},
+			"entity_selectors": getEntitySelectorsSchema(),
+			"filter_expression": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"group_by_fields": {
+				Type:     schema.TypeList,
+				Optional: true,
+				Elem: &schema.Schema{
+					Type: schema.TypeString,
+				},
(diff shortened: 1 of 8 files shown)


sumologic/resource_sumologic_cse_first_seen_rule.go
```diff
@@ -0,0 +1,213 @@
...
+		Schema: map[string]*schema.Schema{
+			"baseline_type": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"baseline_window_size": {
+				Type:     schema.TypeString,
+				Optional: true,
+			},
+			"category": {
```
josh-williams, line 28: "category" is sort of deprecated at the API layer, and auto-calculated based on the tags provided, so we haven't exposed it in other TF rule resources. So I think we can remove it here as well.

sumologic/resource_sumologic_cse_first_seen_rule.go
```diff
@@ -0,0 +1,213 @@
...
+		},
+
+		Schema: map[string]*schema.Schema{
+			"baseline_type": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"baseline_window_size": {
+				Type:     schema.TypeString,
+				Optional: true,
```
josh-williams, line 26: Shouldn't this be required? It is in the OpenAPI spec.

sumologic/resource_sumologic_cse_first_seen_rule.go
```diff
@@ -0,0 +1,213 @@
...
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"name_expression": {
+				Type:     schema.TypeString,
+				Required: true,
+			},
+			"retention_window_size": {
+				Type:     schema.TypeString,
+				Optional: true,
```
josh-williams, line 66: Shouldn't this be required? It is in the OpenAPI spec.

website/docs/r/cse_first_seen_rule.html.markdown
```diff
@@ -0,0 +1,72 @@
...
+  value_fields          = ["dstDevice_hostname"]
+}
+```
+
+## Argument reference
+
+The following arguments are supported:
+
+- `baseline_type` - (Required) The baseline type. Current acceptable values are GLOBAL or PER_ENTITY
+- `baseline_window_size` - (Optional) The baseline window size in milliseconds
```
josh-williams, line 45: Now that you've made the window sizes required and removed the category in the code, we need to do it in the doc too.

sumologic/resource_sumologic_cse_first_seen_rule.go
```diff
@@ -0,0 +1,207 @@
...
+
+	if CSEFirstSeenRuleGet == nil {
+		log.Printf("[WARN] CSE FirstSeen Rule not found, removing from state: %v - %v", id, err)
+		d.SetId("")
+		return nil
+	}
+
+	d.Set("baseline_type", CSEFirstSeenRuleGet.BaselineType)
+	d.Set("baseline_window_size", CSEFirstSeenRuleGet.BaselineWindowSize)
+	d.Set("category", CSEFirstSeenRuleGet.Category)
```
josh-williams, line 111: category should be removed here right?

sumologic/sumologic_cse_first_seen_rule.go
```diff
@@ -0,0 +1,95 @@
...
+type CSEFirstSeenRuleResponse struct {
+	CSEFirstSeenRule CSEFirstSeenRule `json:"data"`
+}
+
+type CSEFirstSeenRule struct {
+	ID                    string           `json:"id,omitempty"`
+	AssetField            string           `json:"assetField"`
+	BaselineType          string           `json:"baselineType"`
+	BaselineWindowSize    string           `json:"baselineWindowSize"`
+	Category              string           `json:"category"`
```
josh-williams, line 80: category should be removed here right?

josh-williams (commented):
LGTM other than these few remaining comments.

PR #475, opened 2023-01-23
SUMO-201782: remove regex check from terrafrom partition data dier field

ksbagr:
https://github.com/Sanyaku/sumologic/pull/116697

We have moved the dataTier validation to the API layer to validate which tier are allowed for a customer based on if they are flex customer or not. So removing the regex checks form the terraform provider. Also updated desc for analyticsTier field in partition.

### DIFF
--- a/sumologic/resource_sumologic_partition.go
+++ b/sumologic/resource_sumologic_partition.go
@@ -30,8 +30,8 @@ func resourceSumologicPartition() *schema.Resource {
 				ValidateFunc: validation.StringLenBetween(1, 16384),
 			},
 			"analytics_tier": {
-				Type:         schema.TypeString,
-				Optional:     true,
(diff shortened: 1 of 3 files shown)


AyanGhatak (approved):

sumovishal (changes requested):
Tests are failing...

=== RUN   TestAccSumoLogicPartition_basic
    testing.go:705: Step 0 error: errors during apply:
        
        Error: ***"id":"FX7EY-RJV3R-N9T5Q","errors":[***"code":"data_streams:invalid_analytics_tier","message":"The Analytics Tier is invalid","detail":"Possible values of Analytics Tier are continuous,frequent,infrequent","meta":***"reason":"Unknown AnalyticsTier "***]***

PR #482, opened 2023-02-10
SUMO-207948, Add support to customize resolution payload on monitors

rmamilla-sumo:
Custom resolution payload feature is already GA'ed and this PR is for corresponding TF changes.

### DIFF
--- a/sumologic/resource_sumologic_connection.go
+++ b/sumologic/resource_sumologic_connection.go
@@ -213,10 +213,7 @@ func resourceToConnection(d *schema.ResourceData) Connection {
 	connection.Headers = mapToHeaders(d.Get("headers").(map[string]interface{}))
 	connection.CustomHeaders = mapToHeaders(d.Get("custom_headers").(map[string]interface{}))
 	connection.DefaultPayload = d.Get("default_payload").(string)
-	if d.Get("resolution_payload").(string) != "" {
-		v := d.Get("resolution_payload").(string)
-		connection.ResolutionPayload = &v
-	}
+	connection.ResolutionPayload = d.Get("resolution_payload").(string)
 	connection.WebhookType = d.Get("webhook_type").(string)
 	connection.ConnectionSubtype = d.Get("connection_subtype").(string)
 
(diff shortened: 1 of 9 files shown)


rmamilla-sumo:
Monitor resource related grouping fields tests are failing due to a bug introduced recently. Fix is already in long - https://imcty.kumoroku.com/?id=119791&assemblies=monitor-manager%2C+api%2C+service&repo=sumo

sumologic/sumologic_client.go
```diff
@@ -447,6 +447,7 @@ type Connection struct {
 	DefaultPayload    string    `json:"defaultPayload"`
 	WebhookType       string    `json:"webhookType"`
 	ConnectionSubtype string    `json:"connectionSubtype,omitempty"`
+	ResolutionPayload *string   `json:"resolutionPayload"`
```
rmamilla-sumo, line 450: Made resolution payload a string pointer instead of string because, i didn't see another option to send null for resolution payload unless specifying string pointer in go.

sumologic/resource_sumologic_connection.go
```diff
@@ -85,6 +85,12 @@ func resourceSumologicConnection() *schema.Resource {
 				Optional:     true,
 				ValidateFunc: validation.StringInSlice([]string{"Incident", "Event"}, false),
 			},
+			"resolution_payload": {
+				Type:         schema.TypeString,
+				Optional:     true,
+				ValidateFunc: validation.StringIsJSON,
+				Computed:     true,
```
rmamilla-sumo, line 92: Added computed:true for resolution payload because, if nothing is set by user, a default value is set on server side and returned.

gchairuangsang (approved):
lgtm

sumologic/resource_sumologic_connection.go
tarunk2: If user specified value is different from the server-side value during the update, does `terraform apply` apply the new change for this field?

rmamilla-sumo: Yes. it will apply that value

PR #489, opened 2023-02-27
finalize CHANGELOG for v2.20.1

dagould:


sumovishal:
This should a minor release `2.21.0` as we have 2 new resources

PR #496, opened 2023-03-09
Check for errors before logging the response.

mflis-sumo:
When there was an HTTP error (e.g. timeout), we would get an error like this, because we read fields from `resp` that was nil:

```
panic: runtime error: invalid memory address or nil pointer dereference
```


sumovishal (approved):
Thanks for fixing this. :)

PR #498, opened 2023-03-10
SUMO-211620 Reduce meaningless diff for sumologic_content.

mflis-sumo:
Perform normalization of json string under
the "config" field to remove some of the logically equivalent changes in terraform diff output:
e.g. absent map entry vs map entry with a null value or absent map entry vs map entry with a default value

### DIFF
--- a/sumologic/resource_sumologic_content.go
+++ b/sumologic/resource_sumologic_content.go
@@ -3,6 +3,7 @@ package sumologic
 import (
 	"encoding/json"
 	"log"
+	"strings"
 	"time"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/schema"
@@ -28,6 +29,7 @@ func resourceSumologicContent() *schema.Resource {
 				ValidateFunc:     validation.StringIsJSON,
 				Required:         true,
 				DiffSuppressFunc: structure.SuppressJsonDiff,
+				StateFunc:        configStateFunc,
 			},
 		},
 		Timeouts: &schema.ResourceTimeout{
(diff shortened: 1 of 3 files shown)


sumovishal:
Let's wait till we have fixed the issue with Content Import/Export API.

mflis-sumo:
@sumovishal is it fixed now?

sumologic/resource_sumologic_content.go
```diff
@@ -39,6 +41,73 @@ func resourceSumologicContent() *schema.Resource {
...
+					for _, queryInterface := range queries {
+						queryItem := queryInterface.(map[string]interface{})
+
+						if queryItem["outputCardinalityLimit"] == nil {
+							queryItem["outputCardinalityLimit"] = 1000
+						}
+
+						if queryItem["transient"] == nil {
+							queryItem["transient"] = false
+						}
```
sumovishal, line 103: How are these used?

sumovishal (commented):
Mostly okay, have one clarifying comment.

sumologic/resource_sumologic_content.go
mflis-sumo: Values of these fields (e.g. `outputCardinalityLimit`) are often unspecified in terraform. JSON returned from the Sumologic API contains these fields with default values (e.g. `"outputCardinalityLimit": 1000`). As a result we get a lot of diff outputs like this:
```
                         ~ {
                                id                                     = null
                              ~ queries                                = [
                                  ~ {
                                      - outputCardinalityLimit = 1000
                                      - spansQueryData         = null
                                      - transient              = false
                                        # (8 unchanged attributes hidden)
                                    },
                                ]
                                # (9 unchanged attributes hidden)
```
My aim here was to get rid of this diff that doesn't bring any value.

sumovishal (approved):

PR #507, opened 2023-03-23
Delete resource_sumologic_polling_source

vsinghal13:


maimaisie:
Is it safe to delete?

sumovishal:
We should not delete resource right away. See this guide for removing a resource https://developer.hashicorp.com/terraform/plugin/sdkv2/best-practices/deprecations#provider-data-source-or-resource-removal

PR #508, opened 2023-04-10
Suppress partition resource changed for TF for total_bytes prop

atishya-22:
Always return `true` to ignore TF resource change(`terraform plan`) check on total_bytes field as it will keep changing with data ingested in partition.

### DIFF
--- a/sumologic/resource_sumologic_partition.go
+++ b/sumologic/resource_sumologic_partition.go
@@ -50,6 +50,9 @@ func resourceSumologicPartition() *schema.Resource {
 			"total_bytes": {
 				Type:     schema.TypeInt,
 				Computed: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					return true
+				},
 			},
 			"data_forwarding_id": {
 				Type:     schema.TypeString,


sumologic/resource_sumologic_partition.go
```diff
@@ -50,6 +50,9 @@ func resourceSumologicPartition() *schema.Resource {
 			"total_bytes": {
 				Type:     schema.TypeInt,
 				Computed: true,
+				DiffSuppressFunc: func(k, old, new string, d *schema.ResourceData) bool {
+					return true
+				},
```
atishya-22, line 55: @dagould @sumovishal What we are trying to achieve here is to not show that resource has changed during `terraform plan` if attribute `total_bytes` has changed. 
I could see that always returning `true` using `DiffSuppressFunc` can be used to suppress this but not sure if there's better alternative that can be done. 
Is there a way to smh ignore this during the serialization of the resource itself? 

I can see some other alternatives like [ignore_changes](https://developer.hashicorp.com/terraform/language/meta-arguments/lifecycle#ignore_changes), [customizeDiff](https://developer.hashicorp.com/terraform/plugin/sdkv2/resources/customizing-differences) but these doesn’t look like any different than to directly have inside resource schema.

ErikAtSumo:
I'm looking to clean up some old PRs, and this one seems like an easy, obvious improvement. @atishya-22 @sumovishal Can you think of any reason to NOT merge this as-is?

ErikAtSumo (approved):

PR #514, opened 2023-05-04
SUMO-218195 added tags support for SLO

jinkane:


website/docs/r/slo.html.markdown
```diff
@@ -154,6 +170,7 @@ The following arguments are supported:
   , `Other`. Defaults to `Latency`.
 - `service` - (Optional) Name of the service.
 - `application` - (Optional) Name of the application.
+- `tags` - (Optional) A map defining tag keys and tag values for the SLO.
```
apoorv-garg, line 173: can we add the beta notice here, like we did for multi-window multi-burn

apoorv-garg (commented):
Have we tested SLOs with previous version of TF provider are not affected upon upgrading to this version of the provider ? (That is, plan stage output is no-op)

sumologic/resource_sumologic_slo_library_slo_test.go
```diff
@@ -0,0 +1,418 @@
...
+  signal_type = "Error"
+  service     = "auth"
+  application = "login"
+  compliance {
+      compliance_type = "Rolling"
+      size            = "1d"
+      target          = 95
+      timezone        = "Asia/Kolkata"
+  }
+  tags = {
```
apoorv-garg, line 152: I'm not sure if these tests will pass, because the PR builder runs the tests against some org in us2 (I think) and we haven't enabled tags there

apoorv-garg: We can check which org it is, and enable them

website/docs/r/slo.html.markdown
jinkane: added for `tags`, removed the same for `burn_rate`

PR #534, opened 2023-06-21
updated query parameter description

ksbagr:
SUMO-223383 - sumo internal

### DIFF
--- a/website/docs/r/log_search.html.markdown
+++ b/website/docs/r/log_search.html.markdown
@@ -29,6 +29,19 @@ resource "sumologic_log_search" "example_log_search" {
             }
         }
     }
+    
+    query_parameter {
+        name        = "my_param1"
+        description = "I am a query parameter. Fill me in."
+        dataType    = "ANY"
+        value       = "*"
+    }
+    query_parameter {
+        name        = "my_param2"
+        description = "I am also query parameter. Fill me in too."
+        dataType    = "NUMBER"
+        value       = "2"
+    }
 
     schedule {
         cron_expression = "0 0 * * * ? *"
@@ -73,31 +86,9 @@ The following arguments are supported:
 - `description` - (Optional) Description of the search.
 - `parent_id` - (Required) The identifier of the folder to create the log search in.
 - `query_string` - (Required) Log query to perform.
-- `query_parameter` - (Optional) List of objects with schema as mentioned below. Values for search template used in the search query. 
+- `query_parameter` - (Block List, Optional) List of query parameters for the log search. See [query parameter schema](#schema-for-query_parameter).
+    Values for search template used in the search query. 
     Learn more about the search templates here : https://help.sumologic.com/docs/search/get-started-with-search/build-search/search-templates/
-    Schema for the objects:
-        - `name` - (Required) The name of the parameter.
-        - `description` - (Optional) A description of the parameter.
-        - `dataType` - (Required) The data type of the parameter. Supported values are:
-            1. `NUMBER`
-            2. `STRING`
-            3. `ANY`
-            4. `KEYWORD`
-        - `value` - (Required) A value for the parameter. Should be compatible with the type set in dataType field.
-    Example:
-        query_parameter {
-            name        = "my_param1"
-            description = "I am a query parameter. Fill me in."
-            dataType    = "ANY"
-            value       = "*"
-        }
-        query_parameter {
-            name        = "my_param2"
-            description = "I am also query parameter. Fill me in too."
-            dataType    = "NUMBER"
-            value       = "2"
-        }
-
 - `parsing_mode` - (Optional) Define the parsing mode to scan the JSON format log messages. Possible values are:
     `AutoParse` and  `Manual`. Default value is `Manual`.
 
@@ -108,6 +99,16 @@ The following arguments are supported:
 - `schedule` - (Block List, Max: 1, Optional) Schedule of the log search. See [schedule schema](#schema-for-schedule)
 - `run_by_receipt_time` - (Optional) This has the value `true` if the search is to be run by receipt time and
     `false` if it is to be run by message time. Default value is `false`.
+
+### Schema for `query_parameter`
+- `name` - (Required) The name of the parameter.
+- `description` - (Optional) A description of the parameter.
+- `dataType` - (Required) The data type of the parameter. Supported values are:
+  1. `NUMBER`
+  2. `STRING`
+  3. `ANY`
+  4. `KEYWORD`
+- `value` - (Required) A value for the parameter. Should be compatible with the type set in dataType field.
 
 
 ### Schema for `schedule`
--- a/website/docs/r/log_search.html.markdown
+++ b/website/docs/r/log_search.html.markdown
@@ -73,9 +73,9 @@ The following arguments are supported:
 - `description` - (Optional) Description of the search.
 - `parent_id` - (Required) The identifier of the folder to create the log search in.
 - `query_string` - (Required) Log query to perform.
-- `query_parameters` - (Optional) Array of objects (LogSearchQueryParameterSyncDefinitionBase). Values for search template used in the search query. 
+- `query_parameter` - (Optional) List of objects with schema as mentioned below. Values for search template used in the search query. 
     Learn more about the search templates here : https://help.sumologic.com/docs/search/get-started-with-search/build-search/search-templates/
-    Schema for `LogSearchQueryParameterSyncDefinitionBase`
+    Schema for the objects:
         - `name` - (Required) The name of the parameter.
         - `description` - (Optional) A description of the parameter.
         - `dataType` - (Required) The data type of the parameter. Supported values are:
@@ -83,7 +83,21 @@ The following arguments are supported:
             2. `STRING`
             3. `ANY`
             4. `KEYWORD`
-        - `value` - A value for the parameter. Should be compatible with the type set in dataType field.
+        - `value` - (Required) A value for the parameter. Should be compatible with the type set in dataType field.
+    Example:
+        query_parameter {
+            name        = "my_param1"
+            description = "I am a query parameter. Fill me in."
+            dataType    = "ANY"
+            value       = "*"
+        }
+        query_parameter {
+            name        = "my_param2"
+            description = "I am also query parameter. Fill me in too."
+            dataType    = "NUMBER"
+            value       = "2"
+        }
+
 - `parsing_mode` - (Optional) Define the parsing mode to scan the JSON format log messages. Possible values are:
     `AutoParse` and  `Manual`. Default value is `Manual`.
 
--- a/website/docs/r/log_search.html.markdown
+++ b/website/docs/r/log_search.html.markdown
@@ -16,7 +16,7 @@ resource "sumologic_log_search" "example_log_search" {
     name = "Demo Search"
     description = "Demo search description"
     parent_id = data.sumologic_personal_folder.personalFolder.id
-    query_string = "_sourceCategory=api error | count by _sourceHost"
+    query_string = "_sourceCategory=api | parse \"auth=*:*:*:*:*,\" as a,email,cid,uid,x | parse \"[logger=*,\" as loggerx | where loggerx matches {{logger}} | where cid matches {{customerID}} | count by _sourceHost"
     parsing_mode =  "AutoParse"
     run_by_receipt_time = true
 
@@ -31,16 +31,16 @@ resource "sumologic_log_search" "example_log_search" {
     }
     
     query_parameter {
-        name        = "my_param1"
-        description = "I am a query parameter. Fill me in."
-        dataType    = "ANY"
-        value       = "*"
+        name          = "logger"
+        description   = "The logger for which the result will be returned"
+        data_type     = "STRING"
+        value = "*"
     }
     query_parameter {
-        name        = "my_param2"
-        description = "I am also query parameter. Fill me in too."
-        dataType    = "NUMBER"
-        value       = "2"
+        name          = "customerID"
+        description   = "The customer id for which the result will be returned"
+        data_type     = "STRING"
+        value = "*"
     }
 
     schedule {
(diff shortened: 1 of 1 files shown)


website/docs/r/log_search.html.markdown
```diff
@@ -73,7 +73,8 @@ The following arguments are supported:
 - `description` - (Optional) Description of the search.
 - `parent_id` - (Required) The identifier of the folder to create the log search in.
 - `query_string` - (Required) Log query to perform.
-- `query_parameters` - (Optional) TODO Find a good description.
+- `query_parameters` - (Optional) Values for search template used in the search query.
```
janek-sumo, line 76: What is an example of this in terraform? I assume this is a complex type. Does this nest blocks, is it a list? As a user I don't see examples of this in the documentation, so it would be helpful to know the schema of this argument.

website/docs/r/log_search.html.markdown
```diff
@@ -73,7 +73,17 @@ The following arguments are supported:
...
+    Learn more about the search templates here : https://help.sumologic.com/docs/search/get-started-with-search/build-search/search-templates/
+    Schema for `LogSearchQueryParameterSyncDefinitionBase`
+        - `name` - (Required) The name of the parameter.
+        - `description` - (Optional) A description of the parameter.
+        - `dataType` - (Required) The data type of the parameter. Supported values are:
+            1. `NUMBER`
+            2. `STRING`
+            3. `ANY`
+            4. `KEYWORD`
+        - `value` - A value for the parameter. Should be compatible with the type set in dataType field.
```
janek-sumo, line 86: I think you should added `(Required)` here too, since you did that for the other arguments.

website/docs/r/log_search.html.markdown
```diff
@@ -73,7 +73,17 @@ The following arguments are supported:
 - `description` - (Optional) Description of the search.
 - `parent_id` - (Required) The identifier of the folder to create the log search in.
 - `query_string` - (Required) Log query to perform.
-- `query_parameters` - (Optional) TODO Find a good description.
+- `query_parameters` - (Optional) Array of objects (LogSearchQueryParameterSyncDefinitionBase). Values for search template used in the search query.
```
janek-sumo, line 76: Have you tested that this works? This is my first time looking at the code, but it seems like the argument name is `query_paramenter`, not `query_parameters`.
https://github.com/SumoLogic/terraform-provider-sumologic/blob/master/sumologic/resource_sumologic_log_search.go#L67

janek-sumo: Could we also add an example in the examples above?

janek-sumo: I think this would be similar to how we document `schedule` or `time_range`. In other words, it's a block list (with a max of 10, according to the docs you linked, but not enforced in the provider code 😞) where we can define multiple `query_parameter` blocks, e.g.

```
query_parameter {
  name        = "my_param1"
  description = "I am a query parameter. Fill me in."
  dataType    = "ANY"
  value       = "*"
}

query_parameter {
  name        = "my_param2"
  description = "I am also query parameter. Fill me in too."
  dataType    = "ANY"
  value       = "*"
}
```

I'm also not sure how `value` is supposed to support the same use case we see in the docs and UI ("Set Values for Parameter").

website/docs/r/log_search.html.markdown
```diff
@@ -29,6 +29,19 @@ resource "sumologic_log_search" "example_log_search" {
             }
         }
     }
+    
+    query_parameter {
```
janek-sumo, line 33: Have you tested this?

For this example to make more sense the `query_string` should use the query parameters defined here.

website/docs/r/log_search.html.markdown
```diff
@@ -85,6 +100,16 @@ The following arguments are supported:
...
 
+### Schema for `query_parameter`
+- `name` - (Required) The name of the parameter.
+- `description` - (Optional) A description of the parameter.
+- `dataType` - (Required) The data type of the parameter. Supported values are:
+  1. `NUMBER`
+  2. `STRING`
+  3. `ANY`
+  4. `KEYWORD`
+- `value` - (Required) A value for the parameter. Should be compatible with the type set in dataType field.
```
janek-sumo, line 111: I don't see how this lines up with the [documentation](https://help.sumologic.com/docs/search/get-started-with-search/build-search/search-templates/). 

Here's what I see when I create a query parameter in the UI.
image

When I select "Manage Parameter Settings" I get this:
image

So at this point does `value` represent the "Default Value" you see in the above screenshot? If not, then I assume it address the "Set Values for Parameter" settings. Here is what that looks like:
image
Multiple options for format:
image
Let's say I select "Lookup". Now how would I go about filling out these options using this Terraform provider?
image

janek-sumo: I did the local testing for you and have results...

This `value` is in fact the "Default Value" you see in the UI. The provider should be updated to rename this field correctly to reflect that, otherwise it's confusing. I recommend `default_value`.

website/docs/r/log_search.html.markdown
janek-sumo: This hasn't been tested, because as-is this does not work. If the `query_parameter`s don't match up with the `query_string` then you'll get this error, which I found when testing this locally:
```
│ Error: {"id":"<removed>","errors":[{"code":"ss:parameter_mismatch","message":"Provided log search is invalid.","detail":"Error creating scheduled search parameters","meta":{"validationProblems":{}}}]}
│ 
│   with sumologic_log_search.example_log_search,
│   on personal.tf line 3, in resource "sumologic_log_search" "example_log_search":
│    3: resource "sumologic_log_search" "example_log_search" {
```

website/docs/r/log_search.html.markdown
```diff
@@ -73,7 +86,9 @@ The following arguments are supported:
 - `description` - (Optional) Description of the search.
 - `parent_id` - (Required) The identifier of the folder to create the log search in.
 - `query_string` - (Required) Log query to perform.
-- `query_parameters` - (Optional) TODO Find a good description.
+- `query_parameter` - (Block List, Optional) List of query parameters for the log search. See [query parameter schema](#schema-for-query_parameter).
```
janek-sumo, line 89: Calling this a "List" could be misleading. When you write out the HCL it doesn't look like a list. The example now shows how it's used, but I'd recommend adding here that this up to 10 `query_parameter` blocks can be added, one for each parameter in the `query_string`.

website/docs/r/log_search.html.markdown
```diff
@@ -73,7 +86,9 @@ The following arguments are supported:
...
+    Values for search template used in the search query.
```
janek-sumo, line 90: This is not a complete sentence and doesn't seem to fit here.

website/docs/r/log_search.html.markdown
```diff
@@ -29,6 +29,19 @@ resource "sumologic_log_search" "example_log_search" {
             }
         }
     }
+    
+    query_parameter {
+        name        = "my_param1"
+        description = "I am a query parameter. Fill me in."
+        dataType    = "ANY"
```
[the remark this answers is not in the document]
ksbagr: done

website/docs/r/log_search.html.markdown
```diff
@@ -29,6 +29,19 @@ resource "sumologic_log_search" "example_log_search" {
...
+        value       = "*"
+    }
+    query_parameter {
+        name        = "my_param2"
+        description = "I am also query parameter. Fill me in too."
+        dataType    = "NUMBER"
```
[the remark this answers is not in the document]
ksbagr: done

website/docs/r/log_search.html.markdown
```diff
@@ -85,6 +100,16 @@ The following arguments are supported:
 - `run_by_receipt_time` - (Optional) This has the value `true` if the search is to be run by receipt time and
     `false` if it is to be run by message time. Default value is `false`.
 
+### Schema for `query_parameter`
+- `name` - (Required) The name of the parameter.
+- `description` - (Optional) A description of the parameter.
+- `dataType` - (Required) The data type of the parameter. Supported values are:
```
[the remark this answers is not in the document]
ksbagr: done

website/docs/r/log_search.html.markdown
ksbagr: The provider code and api expect the parameter name to be value. Have added the details in the description of the parameter.

website/docs/r/log_search.html.markdown
ksbagr: made changes to the description.

website/docs/r/log_search.html.markdown
ksbagr: modified example to include valid query paramters which are being used in search

website/docs/r/log_search.html.markdown
ksbagr: The `Set Values for Parameter ` part is not supported via the log search api so it wont be supported through terraform. The `value` is used as default value. Added this detail in the description for `value` parameter.

website/docs/r/log_search.html.markdown
```diff
@@ -10,13 +10,13 @@ Provides a Sumologic Log Search.
...
-data "sumologic_personal_folder" "personalFolder" {}
+data "sumologic_personal_folder" "personal_folder" {}
 
 resource "sumologic_log_search" "example_log_search" {
     name = "Demo Search"
     description = "Demo search description"
-    parent_id = data.sumologic_personal_folder.personalFolder.id
-    query_string = "_sourceCategory=api error | count by _sourceHost"
+    parent_id = data.sumologic_personal_folder.personal_folder.id
+    query_string = "_sourceCategory=api | parse \"parameter1=*,\" as parameter1 | parse \"parameter2=*,\" as parameter2 | where parameter1 matches {{param1}} | where parameter2 matches {{param2}} | count by _sourceHost"
```
janek-sumo, line 19: This is a nitpick, but for longer query strings I think this is more readable:
```
    query_string = <<QUERY
_sourceCategory=api
| parse \"parameter1=*,\" as parameter1
| parse \"parameter2=*,\" as parameter2
| where parameter1 matches {{param1}}
| where parameter2 matches {{param2}}
| count by _sourceHost
QUERY
```

website/docs/r/log_search.html.markdown
```diff
@@ -86,9 +86,8 @@ The following arguments are supported:
 - `description` - (Optional) Description of the search.
 - `parent_id` - (Required) The identifier of the folder to create the log search in.
 - `query_string` - (Required) Log query to perform.
-- `query_parameter` - (Block List, Optional) List of query parameters for the log search. See [query parameter schema](#schema-for-query_parameter).
-    Values for search template used in the search query. 
-    Learn more about the search templates here : https://help.sumologic.com/docs/search/get-started-with-search/build-search/search-templates/
+- `query_parameter` - (Block List, Optional) Upto 10 query_parameter blocks can be added one for each parameter in the query string.
```
janek-sumo, line 89: "Upto" should be "Up to".

I think you need a hyphen in here:

Up to 10 `query_parameter` blocks can be added - one for each parameter in the `query_string`.

website/docs/r/log_search.html.markdown
```diff
@@ -103,12 +102,12 @@ The following arguments are supported:
...
 - `name` - (Required) The name of the parameter.
 - `description` - (Optional) A description of the parameter.
-- `dataType` - (Required) The data type of the parameter. Supported values are:
+- `data_type` - (Required) The data type of the parameter. Supported values are:
   1. `NUMBER`
   2. `STRING`
   3. `ANY`
   4. `KEYWORD`
-- `value` - (Required) A value for the parameter. Should be compatible with the type set in dataType field.
+- `value` - (Required) The default value for the parameter. Should be compatible with the type set in dataType field.
```
janek-sumo, line 110: "Should be compatible with the type set in dataType field."

-> 

"It should be compatible with the type set in data_type field."

website/docs/r/log_search.html.markdown
janek-sumo: I believe that means that we only support the default format of `None`

image

I think it's worth adding a note in this resource that setting autocomplete values for a query_parameter is not supported at this time.

janek-sumo: It would be good to test if it can be set in the UI after the resource is created with TF, and that TF won't reset those values afterwards. That at least provides somewhat of a workaround and a partial way to define those types of parameters as code.

website/docs/r/log_search.html.markdown
ksbagr: done

website/docs/r/log_search.html.markdown
```diff
@@ -10,13 +10,18 @@ Provides a Sumologic Log Search.
...
-data "sumologic_personal_folder" "personalFolder" {}
+data "sumologic_personal_folder" "personal_folder" {}
 
 resource "sumologic_log_search" "example_log_search" {
     name = "Demo Search"
     description = "Demo search description"
-    parent_id = data.sumologic_personal_folder.personalFolder.id
-    query_string = "_sourceCategory=api error | count by _sourceHost"
+    parent_id = data.sumologic_personal_folder.personal_folder.id
+    query_string = "_sourceCategory=api
```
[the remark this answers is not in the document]
ksbagr: Okay. Tested with Multiline query and updates the doc .

website/docs/r/log_search.html.markdown
ksbagr: Tested this with terraform. The log search was created with the default value as mentinoed in terraform. I was further able to add the dropdown list values for the parameter using UI and on running terraform plan again it does not show any changes after this.

website/docs/r/log_search.html.markdown
```diff
@@ -79,6 +81,15 @@ resource "sumologic_log_search" "example_log_search" {
             threshold_type = "group"
         }
         time_zone = "America/Los_Angeles"
+
+        parameter {
+            name          = "param1"
```
janek-sumo, line 86: Nitpick: 
This is strangely formatted (too much whitespace) and inconsistent with the rest. The same goes for the parameter below.

website/docs/r/log_search.html.markdown
```diff
@@ -10,18 +10,20 @@ Provides a Sumologic Log Search.
 
 ## Example Usage
 ```hcl
-data "sumologic_personal_folder" "personal_folder" {}
+data "sumologic_personal_folder" "personalFolder" {}
```
janek-sumo: Nitpick:

This is inconsistent with the casing used for for the other resources in this example, which all use snake case.

janek-sumo (approved):

sumovishal:
@ksbagr let me know when you have addressed all the comments. I can force merge the PR (the test don't run when we make doc only changes)

janek-sumo (approved):

Issue #195, opened 2021-05-18
Better validation of `sumologic_monitor` in the provider code


mccartney:
There are numerous constraints on what is a valid `sumologic_monitor` configuration. Some of these constraints seem to be enforced only by the Sumo Logic product (API backend) vs. Terraform provider. Thus they are not detected at `terraform plan` stage, rather only they cause failure at `terraform apply`.

## Issue 1
Monitors Resolved and ResolvedCritical time ranges need to be equal, but there's no mention of that in the docs.
Also - if they need to be equal, why do I need to specify this value twice?

```
  triggers  {
    threshold_type = "GreaterThan"
    threshold = 0
    time_range = "-20m"
    trigger_type = "Critical"
[...]
  }
  triggers  {
    threshold_type = "LessThanOrEqual"
    threshold = 0
    time_range = "-30m"
    occurrence_type = "ResultCount"
    detection_method = "StaticCondition"
[...]
  }
```
yields:
```
13:25:22  Error: {"id":"QVP7L-50PLW-K7I5Y","errors":[{"code":"monitors:unsupported_trigger_time_range","message":"Unsupported trigger time range.","detail":"ResolvedCritical trigger must have the same time range as Critical trigger's time range."}]}
```

## Issue 2
```
10:02:49  Error: {"id":"I2CPZ-S1U0I-AJAMO","errors":[{"code":"service.argument.invalid","message":"The request body you've provided is invalid. [Invalid field: triggers[1].thresholdType should be one of the following: 'LessThan', 'GreaterThan', 'LessThanOrEqual', or 'GreaterThanOrEqual'] "}]}
```

## Issue 3
```
09:10:32  Error: {"id":"0SZKJ-UDKVX-IXKTC","errors":[{"code":"monitors:unsupported_trigger_source","message":"Unsupported triggerSource for the given monitorType.","detail":"Unsupported triggerSource for the given monitorType.'AllTimeSeries' and 'AnyTimeSeries' are the valid triggerSource for a metrics monitor."}]}
```

## Issue 4
```
Error: {"id":"EIYJ9-UU73C-EIFAE","errors":[{"code":"monitors:invalid_detection_method","message":"Invalid detection method.","detail":" Critical and ResolvedCritical should have the same detection method."}]}
```

## Issue 5
```
[...]

PR #544, opened 2023-07-04
Improve time_range validation for sumologic_monitor

mflis-sumo:
Insufficient validation of this field was mentioned as "issue 15" in: https://github.com/SumoLogic/terraform-provider-sumologic/issues/195

### DIFF
--- a/sumologic/resource_sumologic_monitors_library_monitor.go
+++ b/sumologic/resource_sumologic_monitors_library_monitor.go
@@ -691,8 +691,8 @@ func getSloBurnRateTimeRangeSchema(triggerType string) *schema.Schema {
 }
 
 var allowedTimeRanges = []string{
-	"5m", "10m", "15m", "30m", "60m", "180m", "360m", "720m", "1440m",
-	"-5m", "-10m", "-15m", "-30m", "-60m", "-180m", "-360m", "-720m", "-1440m",
+	"2m", "5m", "10m", "15m", "30m", "60m", "180m", "360m", "720m", "1440m",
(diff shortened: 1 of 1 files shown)


sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -677,14 +683,22 @@ func getSloBurnRateTimeRangeSchema(triggerType string) *schema.Schema {
...
 	}
 }
 
-var timeRangeValidation = validation.StringMatch(regexp.MustCompile(`^-?(\d)+[smhd]$`), "Time range must be in the format '-?\\d+[smhd]'. Examples: -15m, 1d, etc.")
+var allowedTimeRanges = []string{
+	"5m", "10m", "15m", "30m", "60m", "180m", "360m", "720m", "1440m",
+	"-5m", "-10m", "-15m", "-30m", "-60m", "-180m", "-360m", "-720m", "-1440m",
+	"1h", "3h", "6h", "12h", "24h",
+	"-1h", "-3h", "-6h", "-12h", "-24h",
+	"-1d", "1d",
```
tarunk2, line 698: We allow 2m time range for some customers. Please see `logs.monitor.allowed.time.durations` and `metrics.monitor.allowed.time.durations` properties. That's the reason we don't have a validation for this in Terraform and OpenAPI schema. I see two options:
- Leave it as is.
- Allow 2m and -2m here as well.
#2 is not too bad since if 2m time range is not allowed for a customer, it will fail at run time for them.

mflis-sumo: Ok, I've added 2m and -2m

tarunk2 (approved):

PR #555, opened 2023-07-28
set default value for use_versioned_api param to true

vsinghal13:
Currently, we do not have any default value for `use_versioned_api` parameter. The default value for the optional boolean in goLang is `false`.

Golang bool does not have a way to identify if the value was explicitly set or not . It is not able to identify the difference between if the value is set to `false` or the parameter is not set thereby taking default value as `false`.

When users are setting the parameter to `false`, it is not being passed to backend and therefore the resource is being created with `use_versioned_api` as `true` despite the config block having `false`.

What we want is:
  * if user specifies true => true
  * if user specifies false => false
  * if user does not specify => true (backward compatibility)

This PR converts the parameter to use a *bool and sets the default value as `true` and we are able to achieve the above behavior except for one caveat. Since the parameter has a default value of `true` it gets passed in other path types as well, eg. `CloudWatchPath`, where this parameter is not actionable, resulting in showing that the value needs to be changed when running `terraform plan` every time. Therefore, we have added a custom Diff method to suppress them when needed.

Testing performed:
- [X] able to create a source with `true` and `false` values for `use_versioned_api`.
- [X] able to create a source without any value for `use_versioned_api` and it takes `true` as default.
- [X] verified `ignore_changes` works for other S3-related resources to ignore any changes to this parameter.

### DIFF
--- a/sumologic/resource_sumologic_cloudwatch_source_test.go
+++ b/sumologic/resource_sumologic_cloudwatch_source_test.go
@@ -182,7 +182,13 @@ resource "sumologic_cloudwatch_source" "cloudwatch" {
           	namespace = "AWS/Route53"
           	tags = ["k1=v1"]
 		}
+		use_versioned_api = false
 	  }
+	  lifecycle {
+		ignore_changes = [
+			path["use_versioned_api"]
+		]
+	}
 }
 `, cName, cDescription, cCategory, sName, sDescription, sCategory, testAwsRoleArn)
 }
--- a/sumologic/resource_sumologic_cloudwatch_source_test.go
+++ b/sumologic/resource_sumologic_cloudwatch_source_test.go
@@ -182,13 +182,7 @@ resource "sumologic_cloudwatch_source" "cloudwatch" {
           	namespace = "AWS/Route53"
           	tags = ["k1=v1"]
 		}
-		use_versioned_api = false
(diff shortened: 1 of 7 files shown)


pmontiel-sumo (approved):

sumovishal:
>For this, we can add the block in the tf resource config
>
>lifecycle {
>     ignore_changes = [
>         path["use_versioned_api"],
>     ]
> }

I don't think this is the right approach. This will not be a good experience for customers. Can we write a custom diff method to suppress diff in some cases?

vsinghal13:
@sumovishal Ideally we should not pass the value in cases where we don't want it. That is why we are using a bool pointer to have a tristate boolean property. I tried setting `pathSettings.UseVersionedApi = nil` to skip that parameter from being passed, but it's not honoring that and still passing the default value. Therefore, had to go this route.

vsinghal13:
@sumovishal I have added the custom diff method to suppress changes when needed.

sumovishal:
> @sumovishal I have added the custom diff method to suppress changes when needed.

thanks, can you get it reviewed by someone from the Collection team?

maimaisie (approved):

vsinghal13:
@sumovishal could you please force merge this as unrelated tests are failing?

PR #557, opened 2023-07-31
Allow retrying on 50x HTTP errors

mflis-sumo:
Use default policy of hashicorp/go-retryablehttp

Ref: https://github.com/hashicorp/go-retryablehttp/blob/main/client.go#L469-L511


sumovishal (approved):
lgtm, thanks for doing this :)

PR #591, opened 2023-11-10
INVS-1778: Request custom target column names to determine match list state when appropriate

bgrey-sumo:
Because the API returns the custom column ID instead of the custom column name in the `target_column` value, if the TF user creates a match list using a custom column name (which is allowed by the API), it detects that the match list is constantly out of sync and needs to update the state. 

This change utilizes a new query parameter that allows the caller to request that the custom column name be returned instead of the ID when looking up a match list.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
-## 2.27.1 (Unreleased)
+## 2.27.1 (November 14, 2023)
+BUG FIXES: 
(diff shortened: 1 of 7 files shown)


pmontiel-sumo:
@bgrey-sumo don't forget to update CHANGELOG.md

bgrey-sumo:
> @bgrey-sumo don't forget to update CHANGELOG.md

Thank you sir. I think this is the way I'm supposed to update it but let me know otherwise.

pmontiel-sumo (approved):

PR #614, opened 2024-01-22
INVS-1962-cse-tag-schema-fix

pmontiel-sumo:
One client reported tag_schema resource not working. While investigating saw the reason behind was not specifying `link` for value options. Which were omitted when doing the API call.


bgrey-sumo:
Wouldn't that tag only cause a problem if the `link` or `label` is empty/default? I'm having trouble connecting the dots as to how it results in the behavior we're seeing.

PR #615, opened 2024-01-24
Added changelog entry of bug fixes for GH-610.

kaushik-sumo:
Added bug fixes in 2.28.2 unreleased for GH-610.
Updated logging in response with key to Status from StatusCode because we were logging the value of Status but in print message using key StatusCode.
Added error handler in retrylable http client to log error messages with correct error code.
PR: [PR Link](https://github.com/SumoLogic/terraform-provider-sumologic/pull/610)


dagould (approved):
lgtm

PR #590, opened 2023-11-08
SUMO-227092: Support terraform for azure event hub log source

yyyttthan99:
PR description: 
Added a new resource sumologic_azure_event_hub_log_source in terraform for the new azure source. 

Tests performed: 
1. Manual test by creating a main.tf and using the dev build provider
- Create
Screenshot 2023-11-08 at 11 43 59 AM

- Update
Screenshot 2023-11-09 at 10 30 56 AM

- Delete
Screenshot 2023-11-09 at 10 35 25 AM
Screenshot 2023-11-09 at 10 35 34 AM

NOTE: There is a TODO item in `azure_event_hub_log_source.html.markdown`. We should add a page for Azure Event Hub Log Source under the section https://help.sumologic.com/03Send-Data/Sources/02Sources-for-Hosted-Collectors/ . Then we need to update this doc once we have that page.

PR #626, opened 2024-02-23
SUMO-227092: Terraform support for azure source

yyyttthan99:
Addressed comments in [#590](https://github.com/SumoLogic/terraform-provider-sumologic/pull/590). Closing the last pr since the base branch is too old. 

I was able to create the source with the dev build and acceptance tests can pass now.


website/docs/r/azure_event_hub_log_source.html.markdown
```diff
@@ -0,0 +1,54 @@
...
+    type = "AzureEventHubAuthentication"
+	  shared_access_policy_name = "%s"
+	  shared_access_policy_key = "%s"
+  }
+  path {
+    type = "AzureEventHubPath"
+	  namespace     = "%s"
+	  event_hub_name = "%s"
+	  consumer_group = "%s"
+    region = "%s"
```
maimaisie, line 31: indentation is off on some lines

sumologic/resource_sumologic_generic_polling_source.go
```diff
@@ -315,6 +343,10 @@ func resourceToGenericPollingSource(d *schema.ResourceData) (PollingSource, erro
 		URL:          d.Get("url").(string),
 	}
 
+	if source.ContentType == "AzureEventHubLog" {
+		pollingSource.ScanInterval = 300000
```
maimaisie, line 347: are we setting it because this is a required field in TF? it's optional in Sumo so I wasn't expecting we have to set it explicitly

sumologic/resource_sumologic_generic_polling_source.go
```diff
@@ -342,17 +374,26 @@ func getPollingThirdPartyPathAttributes(pollingResource []PollingResource) []map
...
+			"custom_services":     flattenCustomServices(t.Path.CustomServices),
+			"tag_filters":         flattenPollingTagFilters(t.Path.TagFilters),
+			// "sns_topic_or_subscription_arn": flattenPollingSnsTopicOrSubscriptionArn(t.Path.SnsTopicOrSubscriptionArn),
+			"namespace":      t.Path.Namespace,
+			"event_hub_name": t.Path.EventHubName,
+			"consumer_group": t.Path.ConsumerGroup,
+			"region":         t.Path.Region,
+		}
+
+		if t.Path.SnsTopicOrSubscriptionArn != nil {
```
maimaisie, line 393: can you give some context to the changes on `SnsTopicOrSubscriptionArn` field in this PR? I see changes in a few other places for this field but not sure why we need them here

yyyttthan99: I think I added this while debugging an exception in the acc test which indicates no changes should occur after applying the terraform plan.  This turned out to be fixed by adding  
```
lifecycle {
   ignore_changes = [authentication[0].private_key]
} 
```
Also I tried to eliminate the field after running terraform plan. But seems we cannot avoid that even with current change. We can remove these changes. 

Screenshot 2024-04-25 at 3 58 38 PM

sumologic/resource_sumologic_generic_polling_source.go
yyyttthan99: We will encounter this error even we have already set the scan_interval as an optional field. 
Screenshot 2024-04-25 at 4 45 02 PM

website/docs/r/azure_event_hub_log_source.html.markdown
yyyttthan99: This is strange. Looks good on my end
Screenshot 2024-04-30 at 10 06 33 AM

sumologic/resource_sumologic_generic_polling_source.go
maimaisie: That means it's sending 0 when creating the source, but I think TF default value for `scan_interval` should match that of Sumo API

samjsong: What do other sources do for scan_interval? I'm not a TF expert, why is an optional field giving a default value in TF?

yyyttthan99: Linking slack thread: https://sumologic.slack.com/archives/CAU3N0Y92/p1714602126756289

sumologic/resource_sumologic_generic_polling_source.go
```diff
@@ -25,18 +25,19 @@ func resourceSumologicGenericPollingSource() *schema.Resource {
 		Required: true,
 		ForceNew: true,
 		ValidateFunc: validation.StringInSlice([]string{"AwsS3Bucket", "AwsElbBucket", "AwsCloudFrontBucket",
-			"AwsCloudTrailBucket", "AwsS3AuditBucket", "AwsCloudWatch", "AwsInventory", "AwsXRay", "GcpMetrics", "AwsS3ArchiveBucket"}, false),
+			"AwsCloudTrailBucket", "AwsS3AuditBucket", "AwsCloudWatch", "AwsInventory", "AwsXRay", "GcpMetrics", "AwsS3ArchiveBucket", "AzureEventHubLog"}, false),
 	}
 	pollingSource.Schema["scan_interval"] = &schema.Schema{
 		Type:     schema.TypeInt,
-		Required: true,
+		Optional: true,
```
yyyttthan99, line 32: Need to provide a default value.

sumologic/resource_sumologic_generic_polling_source.go
```diff
@@ -25,18 +25,19 @@ func resourceSumologicGenericPollingSource() *schema.Resource {
...
 	}
 	pollingSource.Schema["paused"] = &schema.Schema{
 		Type:     schema.TypeBool,
-		Required: true,
+		Optional: true,
 	}
 	pollingSource.Schema["url"] = &schema.Schema{
 		Type:     schema.TypeString,
+		Optional: true,
 		Computed: true,
```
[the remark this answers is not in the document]
sumovishal: A computed field which is optional can be set by users. If you don't want users to set the field, don't make it optional.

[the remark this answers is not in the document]
yyyttthan99: Thanks @sumovishal @vsinghal13 , we won't change it in this case.

yyyttthan99:
Thanks Vijit for bringing up the Sensitive field [here](https://sumologic.slack.com/archives/CAU3N0Y92/p1714760010536999) regarding the issue with password field.  I tried that one but unfortunately even with Sensitive to true, it doesn't function properly and we will still have the same issue. Existing fields like secret_key in S3 and private_key in GCP metrics are all using plain text, so I think this behavior is acceptable for share_access_policy_key as well.     

We had discussions about cutoffTimestamp in [slack](https://sumologic.slack.com/archives/C02QNUDUDFW/p1715035885024809) and weekly sync. To summarize, with current changes, user can only update cutoffTimestamp to a non-zero value with tf, this is a discrepancy in tf that we have to live with. If users do want to update cutoffTimestamp to 0, they can do it via API, which we believe is a very rare case and to achieve "Collection from Now", user can alternatively just give the exact epoch timestamp of Now. And we have confirmed ingestion won't be affected.

sumologic/resource_sumologic_generic_polling_source.go
maimaisie: I thought by keeping the hardcoded setting for event hub source we wouldn't need to change the scan interval from required to optional? https://github.com/SumoLogic/terraform-provider-sumologic/pull/626/files#diff-3cd6cf5332a73badbeed5ade27b5168f076cd22222dd45f273384d746dfb1c21R349

yyyttthan99: Yeah even we keep the hardcoded setting, we have to make it optional otherwise customers are required to set scan_interval for azure event hub log sources. 
Screenshot 2024-05-09 at 8 39 13 PM

maimaisie (approved):
Can you check with Damo in #collection-rwc-dev-qe to see if QE needs to do testing on new TF source? If needed can they squeeze this in for S2?

yyyttthan99:
Can I get one more approval for this PR? Thanks! cc @samjsong @vsinghal13 @yuting-liu

sumologic/resource_sumologic_generic_polling_source.go
yuting-liu: Any need to modify the polling source's terraform documentation while we change `scan_interval` to optional?

yuting-liu (approved):
LGTM overall. One small question

yyyttthan99:
>Any need to modify the polling source's terraform documentation while we change scan_interval to optional?

Thanks Yuting for reviewing. Even though we make the scan_interval optional, I am inclined to not exposing this change to customer. The reason is that this change is only for us to handle scan_interval properly, customers don't need to do any changes for azure sources or their existing sources (s3 etc). Besides, I feel it's good practice for the customer to specify scan_interval explicitly. Wdyt?

PR #645, opened 2024-04-18
INVS-2114: Fix match list update function

pdziedzic-sumo:
INVS-2114: Fix match list update function - prevent calling API for unchanged match list items. Add couple of debug logs.
Until now we would call Match List Item update API for every existing item in a Match List. This will be unnecessarily heavy for updates that don't modify these items (i.e. when we add new ones or change match list description only).
Also - couple debug logs will help further identifying the potential reasons for timeout should they still occur.

### DIFF
--- a/sumologic/resource_sumologic_cse_match_list.go
+++ b/sumologic/resource_sumologic_cse_match_list.go
@@ -242,13 +242,22 @@ func resourceSumologicCSEMatchListUpdate(d *schema.ResourceData, meta interface{
 		return fmt.Errorf("[ERROR] An error occurred updating match list with id %s, err: %v", d.Id(), err)
 	}
 
+	var deleteItemIds []string
+	var updateItems []CSEMatchListItemPost
+	var addItems []CSEMatchListItemPost
+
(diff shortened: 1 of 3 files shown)


sumologic/resource_sumologic_cse_match_list.go
```diff
@@ -242,45 +242,52 @@ func resourceSumologicCSEMatchListUpdate(d *schema.ResourceData, meta interface{
...
+				updateItems = append(updateItems, newItem)
+			}
 		} else {
-			addItems = append(addItems, newItem)
+			// match list item doesn't exist any more
+			deleteItemIds = append(deleteItemIds, oldItemId)
 		}
 	}
 
+	log.Printf("[DEBUG] Match List update items - to add: %d, to update: %d, to delete: %d", len(addItems), len(updateItems), len(deleteItemIds))
```
pdziedzic-sumo, line 289: This seems to be the approach to produce debug logs in other files.

pmontiel-sumo (approved):

PR #648, opened 2024-04-24
SUMO-233062: Update the dataType in field resource docs.

AyanGhatak:
https://api.sumologic.com/docs/#operation/createField

Fields should not include the dataType property during its creation.


sumovishal (commented):
This change can break TF customers. Can we look at DB to find out if customers are using values other than string? We still won't know if they were created via TF but that will a start.

What about API? Are we changing the API too b/c this is a backwards incompatible change.

sumovishal (approved):

dagould (approved):
lgtm

dagould:
Merged `master` into this branch to re-trigger build; `tfproviderlint` succeeded.

PR #651, opened 2024-05-04
Feat: Add role v2 terraform support

ssharma-sumologic:
Add terraform resource and dataSource for role v2 api.

Tested locally

image


website/docs/d/role_v2.html.markdown
```diff
@@ -0,0 +1,48 @@
...
+
+```hcl
+data "sumologic_role_v2" "that" {
+  id = "1234567890"
+}
+```
+
+A role can be looked up by either `id` or `name`. One of those attributes needs to be specified.
+
+If both `id` and `name` have been specified, `id` takes precedence.
```
dagould, line 29: If both are specified, and they disagree (the only case in which precedence matters), is that not an error?

dagould (approved):
lgtm

website/docs/d/role_v2.html.markdown
ssharma-sumologic: This is the behaviour that's being used by role v1 data source, so am continuing with the same pattern right now. We can change this in future if needed ( IIUC you are prompting to throw an error if both are specified and they point to different responses , sounds fair tbh but since we would need this change in both v1 and v2 role datasource, am keeping this for a future change, not part of the current pr )

PR #653, opened 2024-05-07
Add Support for Smart Alerts

sumovishal:
- Adding support for smart alerts in Monitor resource.
- Using "Normal" for `status` field to avoid drift b/w backend and state. We should deprecate this field and eventually remove it.
- Fixed some warnings and spacing issues.

Would recommend to enable "Hide Whitespace" when reviewing the PR.


website/docs/r/monitor.html.markdown
```diff
@@ -528,6 +572,14 @@ Here is a summary of arguments for each condition type (fields which are not mar
       - `burn_rate_threshold` (Required): The burn rate percentage threshold.
       - `time_range` (Required): The relative time range for the burn rate percentage evaluation.  Accepted format: Optional `-` sign followed by `<number>` followed by a `<time_unit>` character: `s` for seconds, `m` for minutes, `h` for hours, `d` for days. Examples: `30m`, `-12h`.
 
+#### log_anomaly_condition
+  - `field`: The name of the field that the trigger condition will alert on. The trigger could compare the value of specified field with the threshold. If field is not specified, monitor would default to result count instead.
+  - `sensitivity`: The triggering sensitivity of the anomaly model used for this monitor.
+  - `anomaly_detector_type`: The type of anomaly model that will be used for evaluating this monitor. Possible values are: `SpikeDetector` and `Cluster`.
```
rmamilla-sumo, line 578: possible values should be cluster only. spike is backward compatiblity and should not make to tf atleast.

sumologic/resource_sumologic_monitors_library_monitor_test.go
```diff
@@ -139,9 +143,12 @@ func TestAccSumologicMonitorsLibraryMonitor_schemaTriggerConditionValidations(t
 		testName := "terraform_test_invalid_monitor_" + testNameSuffix
 
 		resource.Test(t, resource.TestCase{
-			PreCheck:     func() { testAccPreCheck(t) },
-			Providers:    testAccProviders,
-			CheckDestroy: testAccCheckMonitorsLibraryMonitorDestroy(monitorsLibraryMonitor),
+			PreCheck:  func() { testAccPreCheck(t) },
+			Providers: testAccProviders,
+			CheckDestroy: func() resource.TestCheckFunc {
```
rmamilla-sumo, line 148: for my learning, i understand what you made a change now but how was old code working?

website/docs/r/monitor.html.markdown
sumovishal: Removed from terraform, we should cleanup our API too.

website/docs/r/monitor.html.markdown
```diff
@@ -333,6 +333,49 @@ JSON
...
+  queries {
+      row_id = "A"
+      query = "_sourceCategory=api error | timeslice 5m | count by _sourceHost"
+  }
+
+  trigger_conditions {
+    logs_anomaly_condition {
+      field = "_count"
+      sensitivity = 0.4
+      anomaly_detector_type = "SpikeDetector"
```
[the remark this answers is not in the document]
sumovishal: Fixed

sumologic/resource_sumologic_monitors_library_monitor_test.go
sumovishal: `testAccCheckMonitorsLibraryMonitorDestroy` did take 1 param, however it was unused. I removed it as part of this PR.

sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -611,6 +619,37 @@ var sloBurnRateTriggerConditionSchema = map[string]*schema.Schema{
...
+		Optional:     true,
+		Default:      "Both",
+		ValidateFunc: validation.StringInSlice([]string{"Both", "Up", "Down"}, false),
+	},
+	"anomaly_detector_type": {
+		Type:         schema.TypeString,
+		Required:     true,
+		ValidateFunc: validation.StringInSlice([]string{"Cluster"}, false),
+	},
+	"sensitivity": {
```
rmamilla-sumo, line 638: shouldn't sensitivity have a validate func which fails when user sets the value > 1 ?

sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -611,6 +619,37 @@ var sloBurnRateTriggerConditionSchema = map[string]*schema.Schema{
...
+		Type:     schema.TypeFloat,
+		Optional: true,
+		Default:  0.5,
+	},
+	"critical": nested(false, schemaMap{
```
rmamilla-sumo, line 643: why min_anomaly_count is under critical ?

sumologic/resource_sumologic_monitors_library_monitor.go
rmamilla-sumo: shouldn't sensitivity also be under critical ?

sumovishal: >shouldn't sensitivity have a validate func which fails when user sets the value > 1 ?

hmm, good catch. We are missing this check from API too. We should add it there.

sumovishal: >shouldn't sensitivity also be under critical ?

`direction`, `anomaly_detector_type`, and `sensitivity` are all tied to anomaly model rather than critical trigger. That's why I did not include them under `critical` object.

rmamilla-sumo: can you confirm with Ash/ Ryley once about sensitivity because that is part of inference evaluation if I'm not wrong.

sumologic/resource_sumologic_monitors_library_monitor.go
```diff
@@ -611,6 +619,37 @@ var sloBurnRateTriggerConditionSchema = map[string]*schema.Schema{
...
+		"min_anomaly_count": {
+			Type:     schema.TypeInt,
+			Optional: true,
+			Default:  1,
```
rmamilla-sumo, line 647: adding validation for this would be hard. so we can ignore adding one to this.

rmamilla-sumo (approved):
LGTM

sumologic/resource_sumologic_monitors_library_monitor.go
sumovishal: The API has the same default value. Even if I remove it from here, we will still use 1 as default value.

sumologic/resource_sumologic_monitors_library_monitor.go
sumovishal: I am going to move `sensitivity` to `critical` trigger.

PR #667, opened 2024-06-25
SUMO-242708: fix x/net, protobuf, go-retyablehttp and go-getter

sjain05:


adisuj (approved):
lgtm - has all the necessary upgrades to fix current dependabot vulns

PR #672, opened 2024-07-11
Update field.html.markdown

jeff-d:
correct argument list details, update examples


sumovishal:
Fields don't support data_type. It was recently removed. See https://github.com/SumoLogic/terraform-provider-sumologic/pull/648

AyanGhatak (approved):

PR #674, opened 2024-07-15
Add IsIncludedInDefaultSearch field to partition resource

ssharma-sumologic:
For flex partitions, there is an option to included the partition in default search or not. Added support for this option in partition resource. 

Also added randomness in names of some tests which were failing.

### DIFF
--- a/sumologic/resource_sumologic_cse_network_block_test.go
+++ b/sumologic/resource_sumologic_cse_network_block_test.go
@@ -2,6 +2,8 @@ package sumologic
 
 import (
 	"fmt"
+	"math/rand"
+	"net"
 	"testing"
 
 	"github.com/hashicorp/terraform-plugin-sdk/helper/resource"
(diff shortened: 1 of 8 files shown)


sumologic/resource_sumologic_cse_network_block_test.go
```diff
@@ -112,3 +114,12 @@ func testCheckNetworkBlockValues(networkBlock *CSENetworkBlock, nAddressBlock st
 		return nil
 	}
 }
+
+func generateRandomCIDRBlock() string {
+	ip := make(net.IP, 4)
```
dagould, line 119: Shouldn't the bits after the prefix (i.e., the last 6) be zeroes?

ssharma-sumologic: Makes sense, have set last octet 0.

AbhishekSumo (approved):

dagould (approved):
lgtm

PR #678, opened 2024-08-07
Csre 183 adding data forwarding resource

Aqz2024:
Added S3 data forwarding support to the terraform provider. 

Slack Conversation - [link](https://api.sumologic.com/docs/#operation/createDataForwardingBucket) 
- The tests are failing because of [this issue](https://sumologic.atlassian.net/browse/SUMO-247862?atlOrigin=eyJpIjoiYzA2YjA5YmUxNDQyNDg1YjllYWYwN2Y4ODA2MWM4NjciLCJwIjoiamlyYS1zbGFjay1pbnQifQ), hence we are not adding the test as a part of this PR. We have locally ran the tests and they are passing, so we will add those tests right after the above issue is fixed.

Documentation
- API docs - https://api.sumologic.com/docs/#operation/createDataForwardingBucket
- Data forwarding Documentation - https://help.sumologic.com/docs/manage/data-forwarding/amazon-s3-bucket/


sumologic/resource_sumologic_data_forwarding_test.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+func testAccCheckDataForwardingDestroy() resource.TestCheckFunc {
+	return func(s *terraform.State) error {
+		client := testAccProvider.Meta().(*Client)
+		for _, r := range s.RootModule().Resources {
+			id := r.Primary.ID
+			p, err := client.getDataForwarding(id)
+
+			if err != nil {
```
Aqz2024, line 149: Is this a good practice? The reason why I did it is to check if the resource is no longer there. Doing a get on the id of the resource that does not exist throws an error. So if I am getting that particular error that mean we are good and the resource is gone.

zackprice (approved):

janek-sumo:
Can you add more context to the PR description? References to the Sumo Logic documentation for the resource would be helpful.

namangoya (commented):
Lgtm

namangoya (approved):

PR #688, opened 2024-09-18
Adding TF support for configuring Data Forwarding Rules

namangoya:
Added TF support for data forwarding /rules.

Slack Conversation - [link](https://api.sumologic.com/docs/#operation/createDataForwardingBucket)

- The tests are failing because of [this issue](https://sumologic.atlassian.net/browse/SUMO-247862?atlOrigin=eyJpIjoiYzA2YjA5YmUxNDQyNDg1YjllYWYwN2Y4ODA2MWM4NjciLCJwIjoiamlyYS1zbGFjay1pbnQifQ), hence I am not adding the test as a part of this PR. I have locally ran the tests and they are passing, so we will add those tests right after the above issue is fixed.

Documentation

- API docs - https://api.sumologic.com/docs/#operation/createDataForwardingRule
- Data forwarding Documentation - https://help.sumologic.com/docs/manage/data-forwarding/amazon-s3-bucket/#forward-datato-s3


sumologic/resource_sumologic_data_forwarding_rule.go
```diff
@@ -0,0 +1,167 @@
...
+				Type:         schema.TypeString,
+				Required:     true,
+				ValidateFunc: validation.StringLenBetween(16, 16),
+			},
+			"enabled": {
+				Type:     schema.TypeBool,
+				Optional: true,
+				Default:  true,
+			},
+			"file_prefix": {
```
ssharma-sumologic, line 38: any particular reason we are using file_prefix instead of file_format, since in documentation also file_format is used right ?

ssharma-sumologic: reason is I was thinking having same name will help in reading

sumologic/sumologic_data_forwarding_rule.go
```diff
@@ -0,0 +1,69 @@
...
+
+func (s *Client) DeleteDataForwardingRule(indexId string) error {
+	url := fmt.Sprintf("v1/logsDataForwarding/rules/%s", indexId)
+
+	_, err := s.Delete(url)
+
+	return err
+}
+
+type DataForwardingRule struct {
```
ssharma-sumologic, line 62: again `ID            string `json:"id,omitempty"` seems to be missing here, is this intentional ?

ssharma-sumologic, line 62: nit: should we add omitempty if needed someplace

ssharma-sumologic (approved):
Added some comments, rest lgtm

sumologic/resource_sumologic_data_forwarding_rule.go
namangoya: Nice pick, I thought of keeping it "file_prefix" just for readability .. but I've refactored it back to "file_format" in my recent commit.

sumologic/sumologic_data_forwarding_rule.go
namangoya: thanks, added in my recent commit

sumologic/resource_sumologic_data_forwarding_rule.go
```diff
@@ -0,0 +1,167 @@
...
+	}
+
+	if dataForwardingRule == nil {
+		log.Printf("[WARN] DataForwarding Rule (%s) not found, removing from state", indexId)
+		d.SetId("")
+
+		return nil
+	}
+
+	d.SetId(dataForwardingRule.IndexId)
```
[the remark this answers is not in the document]
namangoya: > is the id of the rule same as the indexId of the partition/scheduledview

No. The 'id' of the rule (created in db as an auto-incremented id) is different from the 'indexId', but both are unique identifiers for the rule, as every Partition/Scheduled view can have at most one Data-Forwarding rule.

> in the documentation it's mentioned that there are two fields in output, indexId and id. 

I can see that the documentation is little confusing, but we are talking about 'indexId' itself there.

sumologic/sumologic_data_forwarding_rule.go
namangoya: I left it intentionally as the 'indexId' also serves the same purpose i.e. uniquely identifying a rule because we can have only 1 data forwarding rule configured for a partition or scheduled view

namangoya:
I've tested the changes manually also and everything is working fine, attaching the [main.txt](https://github.com/user-attachments/files/17063133/main.txt) (rename it to `main.tf`) and test results. 

#### Steps to test 
- Rename the file to `main.tf`
- Copy the `main.tf` in `/terraform-provider-sumologic` directory (i.e. the repo itself), and add populate the variables
- Run `make install-dev`
- Run `terraform init`
- Run `terraform plan`
- Run `terraform apply`

#### Flow of the tests is 

##### Creating resources 
- Run `terraform plan` and verify the plan and then run `terraform apply`
- Creating a partition.
- Creating a data forwarding destination.
- Configuring the data forwarding rule on the above partition.

##### Deleting resources
- Comment out the resource and run `terraform apply`, it will delete all the 3 resources it created above

Creation 1
Creation 2
Deletion

PR #691, opened 2024-09-20
SUMO-248024: 

namangoya:
#### What are we doing?
- Adding `enabled` to the documentation of sumologic_data_forwarding_destination, and adjusted default values.
- Adding `index_id` in sumologic_scheduled_views, and adjusted necessary documentations.

#### Why is this required?
As a part of supporting Data Forwarding Rules creation via TF for Scheduled Views we need `index_id`, because the `id` is coming from the autoview table.

#### Testing
Manual testing has been performed via `main.tf` and we are successfully able to configure a DF Rule for a SV. Attached test results and the test file in the comment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 DEPRECATIONS:
 * resource_sumologic_ingest_budget : Deprecated in favour of `resource_sumologic_ingest_budget_v2`.
 
-## 2.31.5 (September 23, 2024)
+## 2.31.5
 ENHANCEMENTS:
 * Added *index_id* attribute to sumologic_scheduled_view. (GH-691)
 * Added support for configuring sumologic_data_forwarding_rule for sumologic_scheduled_view. (GH-691)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 DEPRECATIONS:
 * resource_sumologic_ingest_budget : Deprecated in favour of `resource_sumologic_ingest_budget_v2`.
 
+## 2.31.5 (September 23, 2024)
+ENHANCEMENTS:
+* Added *index_id* attribute to sumologic_scheduled_view. (GH-691)
+* Added support for configuring sumologic_data_forwarding_rule for sumologic_scheduled_view. (GH-691)
+
 ## 2.31.4 (September 19, 2024)
 * **New Resource:** sumologic_data_forwarding_destination (GH-678)
 * **New Resource:** sumologic_data_forwarding_rule (GH-688)
--- a/sumologic/resource_sumologic_data_forwarding_rule.go
+++ b/sumologic/resource_sumologic_data_forwarding_rule.go
@@ -33,12 +33,10 @@ func resourceSumologicDataForwardingRule() *schema.Resource {
 			"enabled": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  true,
 			},
 			"file_format": {
 				Type:     schema.TypeString,
 				Optional: true,
-				Default:  "{index}_{day}_{hour}_{minute}_{second}",
 			},
 			"payload_schema": {
 				Type:         schema.TypeString,
--- a/sumologic/sumologic_data_forwarding_rule.go
+++ b/sumologic/sumologic_data_forwarding_rule.go
@@ -62,7 +62,7 @@ func (s *Client) DeleteDataForwardingRule(indexId string) error {
 type DataForwardingRule struct {
 	IndexId       string `json:"indexId"`
 	DestinationId string `json:"destinationId"`
-	Enabled       bool   `json:"enabled,omitempty"`
+	Enabled       bool   `json:"enabled"`
 	FileFormat    string `json:"fileFormat,omitempty"`
 	PayloadSchema string `json:"payloadSchema,omitempty"`
 	Format        string `json:"format,omitempty"`
--- a/website/docs/r/data_forwarding_rule.html.markdown
+++ b/website/docs/r/data_forwarding_rule.html.markdown
@@ -58,7 +58,7 @@ The following arguments are supported:
 - `index_id` - (Required) The *id* of the Partition or *index_id* of the Scheduled View the rule applies to.
 - `destination_id` - (Required) The data forwarding destination id.
 - `enabled` - (Optional) True when the data forwarding rule is enabled. Will be treated as _false_ if left blank.
-- `file_format` - (Optional) Specify the path prefix to a directory in the S3 bucket and how to format the file name.
+- `file_format` - (Optional) Specify the path prefix to a directory in the S3 bucket and how to format the file name. For possible values, kindly refer the point 6 in the [documentation](https://help.sumologic.com/docs/manage/data-forwarding/amazon-s3-bucket/#forward-datato-s3).
 - `payload_schema` - (Optional) Schema for the payload. Default value of the payload schema is _allFields_ for scheduled view, and _builtInFields_ for partition.
   _raw_ payloadSchema should be used in conjunction with _text_ format and vice versa.
 - `format` - (Optional) Format of the payload. Default format will be _csv_. 
--- a/website/docs/r/data_forwarding_rule.html.markdown
+++ b/website/docs/r/data_forwarding_rule.html.markdown
@@ -9,21 +9,53 @@ description: |-
 Provider to manage [Sumologic Data Forwarding Rule](https://help.sumologic.com/docs/manage/data-forwarding/amazon-s3-bucket/#forward-datato-s3)
 
 ## Example Usage
+
+For Partitions
 ```hcl
+resource "sumologic_partition" "test_partition" {
+  name               = "testing_rule_partitions"
+  routing_expression = "_sourcecategory=abc/Terraform"
+  is_compliant       = false
+  retention_period   = 30
+  analytics_tier     = "flex"
+}
+
 resource "sumologic_data_forwarding_rule" "example_data_forwarding_rule" {
-    index_id = "00000000024C6155"
-    destination_id = "00000000000732AA"
-    enabled = "true"
-    file_format = "test/{index}/{day}/{hour}/{minute}"
-    payload_schema = "builtInFields"
-    format = "json"
+  index_id = sumologic_partition.test_partition.id
+  destination_id = "00000000000732AA"
+  enabled = true
+  file_format = "test/{index}/{day}/{hour}/{minute}"
+  payload_schema = "builtInFields"
+  format = "json"
+}
+```
+For Scheduled Views
+```hcl
+resource "sumologic_scheduled_view" "failed_connections" {
+  index_name = "failed_connections"
+  query = "_sourceCategory=fire | count"
+  start_time = "2024-09-01T00:00:00Z"
+  retention_period = 1
+  lifecycle {
+    prevent_destroy = true
+    ignore_changes = [index_id]
+  }
+}
+
+resource "sumologic_data_forwarding_rule" "test_rule_sv" {
+  index_id       = sumologic_scheduled_view.failed_connections.index_id
+  destination_id = sumologic_data_forwarding_destination.test_destination.id
+  enabled        = false
+  file_format    = "test/{index}"
+  payload_schema = "raw"
+  format         = "text"
 }
 ```
 ## Argument reference
 
 The following arguments are supported:
 
-- `index_id` - (Required) The _id_ of the Partition or Scheduled View the rule applies to.
+- `index_id` - (Required) The *id* of the Partition or *index_id* of the Scheduled View the rule applies to.
 - `destination_id` - (Required) The data forwarding destination id.
 - `enabled` - (Optional) True when the data forwarding rule is enabled. Will be treated as _false_ if left blank.
 - `file_format` - (Optional) Specify the path prefix to a directory in the S3 bucket and how to format the file name.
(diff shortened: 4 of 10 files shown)


namangoya:
#### Manual Testing has been performed - Success

- Steps to test the `main.tf` file are mentioned [here](https://github.com/SumoLogic/terraform-provider-sumologic/pull/688#issuecomment-2361608784).
- File for manual testing (rename it to main.tf) - [main.txt](https://github.com/user-attachments/files/17088609/main.txt) 
- All 3 resources will be created
Screenshot 2024-09-23 at 10 07 36 AM
Screenshot 2024-09-23 at 10 07 43 AM

sumologic/sumologic_data_forwarding_rule.go
```diff
@@ -62,7 +62,7 @@ func (s *Client) DeleteDataForwardingRule(indexId string) error {
 type DataForwardingRule struct {
 	IndexId       string `json:"indexId"`
 	DestinationId string `json:"destinationId"`
-	Enabled       bool   `json:"enabled,omitempty"`
+	Enabled       bool   `json:"enabled"`
```
raghavxk, line 65: Why are we making this change?  This will send value as `false` instead of not serialising it in json response(which eventually gets treated as `null`)

website/docs/r/data_forwarding_rule.html.markdown
```diff
@@ -9,25 +9,61 @@ description: |-
...
+  retention_period = 1
+  lifecycle {
+    prevent_destroy = true
+    ignore_changes = [index_id]
+  }
+}
+
+resource "sumologic_data_forwarding_rule" "test_rule_sv" {
+  index_id       = sumologic_scheduled_view.failed_connections.index_id
+  destination_id = sumologic_data_forwarding_destination.test_destination.id
```
raghavxk, line 47: What happens when user supplies both `IndexName` and `IndexId` ?

website/docs/r/data_forwarding_rule.html.markdown
```diff
@@ -9,25 +9,61 @@ description: |-
...
 ```
 ## Argument reference
 
 The following arguments are supported:
 
-- `index_id` - (Required) The _id_ of the Partition or Scheduled View the rule applies to.
+- `index_id` - (Required) The *id* of the Partition or *index_id* of the Scheduled View the rule applies to.
 - `destination_id` - (Required) The data forwarding destination id.
-- `enabled` - (Required) True when the data forwarding rule is enabled.
+- `enabled` - (Optional) True when the data forwarding rule is enabled. Will be treated as _false_ if left blank.
```
raghavxk, line 60: Should we add possible values for `file_format` here to aid customers?

namangoya: Not needed, we already have a descriptive documentation ... and that is added on top of this page - [see here](https://github.com/SumoLogic/terraform-provider-sumologic/pull/691/files#diff-92753b9c76a682404721409b68e2d58d2edadcd8e4ae69ef0ffe01cb66ec65ebR9)

website/docs/r/data_forwarding_rule.html.markdown
namangoya: `IndexId` will simply be ignored ... as `IndexId` is never taken into consideration in PUT/POST calls .. please check the API documentation [here](https://api.sumologic.com/docs/#operation/updateScheduledView)

namangoya: We are just using `IndexId` while creating the DF Rule... it is never updated in the PUT call, nor it is ever required to maintain the SV resource's state.

sumologic/sumologic_data_forwarding_rule.go
namangoya: The bad part is that `enabled = false` is treated as null.... and whenever we are trying to set `false`, it is not sent in the API call... 

We are not setting any default for `enabled` .. so it will not be send as `false` in case left blank (see [here](https://github.com/SumoLogic/terraform-provider-sumologic/pull/691/files/138fe92703280402b4d5c5383b44671a5b1bc8ef#diff-a6863bdbf853ea5b139b3ca10ea5fe15699db83c96727384162c8fe87f5385f3L36)).

I've tested this scenario..

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 DEPRECATIONS:
 * resource_sumologic_ingest_budget : Deprecated in favour of `resource_sumologic_ingest_budget_v2`.
 
+## 2.31.5 (September 23, 2024)
```
ashpaliwal, line 5: Please remove date. Date shall be added on the day of release

raghavxk (approved):
lgtm

CHANGELOG.md
namangoya: Sure, done

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 DEPRECATIONS:
 * resource_sumologic_ingest_budget : Deprecated in favour of `resource_sumologic_ingest_budget_v2`.
```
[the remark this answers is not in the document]
namangoya: Tbh, I thought that 3.0.0 is written for some purpose.. and didn't thought of using that, just continued after 2.31.3 ..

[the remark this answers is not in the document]
sumovishal: @dagould are you planning to maintain two versions of the provider? I would be in favor of just continuing `3.x.x` version.

sumologic/resource_sumologic_data_forwarding_rule.go
```diff
@@ -33,12 +33,10 @@ func resourceSumologicDataForwardingRule() *schema.Resource {
 			"enabled": {
 				Type:     schema.TypeBool,
 				Optional: true,
-				Default:  true,
```
sumovishal, line 36: This looks like a change in behavior. A rule will be disabled by default now? This might be a breaking change. The good part is we are scheduled for a major release if this is a breaking change it should be fine. We should document the new behavior.

website/docs/r/data_forwarding_rule.html.markdown
sumovishal: >IndexId will simply be ignored .

This behavior should be documented in the terraform docs.

website/docs/r/data_forwarding_rule.html.markdown
sumovishal: >Not needed

I disagree, we should cover the possible values here. Can't rely on an example to cover to do that.

website/docs/r/scheduled_view.html.markdown
```diff
@@ -44,6 +45,7 @@ The following arguments are supported:
 The following attributes are exported:
 
 - `id` - The internal ID of the scheduled view.
+- `index_id` - The Index ID of the scheduled view. It never updates at any point of time during resource updates, therefore make sure to ignore this via `ignore_changes = [index_id]`.
```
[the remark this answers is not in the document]
namangoya: This is already enforced in the API ([PUT](https://api.sumologic.com/docs/#operation/updateScheduledView) and [POST](https://api.sumologic.com/docs/#operation/createScheduledView) api docs). Here we are just communicating/informing customers as it is a non-updatable field.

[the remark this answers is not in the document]
namangoya: Even if the user passes `index_id`, it will simply be ignored at the API.

sumologic/resource_sumologic_data_forwarding_rule.go
namangoya: Amm, no it's not a breaking change. Because in the [API](https://api.sumologic.com/docs/#operation/createDataForwardingRule), we don't enforce any default value for `enabled`, which means that it'll be treated as `false` if left blank. 

Here, I'm correcting a mistake from my previous PR .. and and bringing it in-line with the API.

website/docs/r/scheduled_view.html.markdown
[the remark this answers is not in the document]
sumovishal: >Even if the user passes index_id, it will simply be ignored at the API.

I don't think this is right behavior. We are silently dropping a field provided by the client. Instead we should return error b/c request is invalid as you are not allowed to change the index id. This is the real issue here.

It should be noted that `ignore_changes` is not strictly required if you don't change the index id which I believe most customer won't do, given we don't support it?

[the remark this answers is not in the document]
namangoya: > It should be noted that ignore_changes is not strictly required if you don't change the index id which I believe most customer won't do, given we don't support it?

Yes, not a strict requirement to add `ignore_changes` .. but recommended, and hence conveying that msg in the documentation here.


> I don't think this is right behavior. We are silently dropping a field provided by the client. Instead we should return error b/c request is invalid as you are not allowed to change the index id. This is the real issue here.

- This will not even reach the API layer as it'll be dropped by OpenAPI translation itself and will not translated into the request payload's Java Object. 
- Moreover, there can be some invalid fields added by the user (by mistake or intentionally), so those are also simply ignored and not translated by OpenAPI... 
- Therefore, we can't really do much with this `index_id` field being ignored.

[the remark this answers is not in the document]
namangoya: > We can follow-up internally on this one. I don't see a reason why OpenAPI would silently drop a field.

Sure, can I know the set of people who can help me here?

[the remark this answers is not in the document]
namangoya: > The language should be updated to reflect that. Currently, it seems we are asking to add ignore_changes rather than a recommendation.

But without `ignore_changes`, we are running into [this issue](https://sumologic.slack.com/archives/CAU3N0Y92/p1726892494120009?thread_ts=1726887269.983899&cid=CAU3N0Y92).. let's continue for this on the slack thread.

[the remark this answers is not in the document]
namangoya: > But without ignore_changes, we are running into [this issue](https://sumologic.slack.com/archives/CAU3N0Y92/p1726892494120009?thread_ts=1726887269.983899&cid=CAU3N0Y92).. let's continue for this on the slack thread.

Issue is resolved with the introduction of [computed](https://github.com/SumoLogic/terraform-provider-sumologic/pull/691/files#diff-4b1847b87c211c5fa22381b241115d5d3d299920c9b492b8ce8d669a4fcf7d41R43) prop for `index_id`.  (see [thread](https://sumologic.slack.com/archives/CAU3N0Y92/p1727231569976409?thread_ts=1726887269.983899&cid=CAU3N0Y92))

I'm updating the documentation that `ignore_changes` is just a recommendation and not a mandatory thing.

website/docs/r/data_forwarding_rule.html.markdown
namangoya: Shall I point to the [original documentation here](https://help.sumologic.com/docs/manage/data-forwarding/amazon-s3-bucket/#forward-datato-s3) (point 6) in that case? 

Possible values for `file_format` are a combination of few placeholders, and adding all of them here will not look good in the TF docs imho.

website/docs/r/data_forwarding_rule.html.markdown
namangoya: Yes, we are documenting it in the SV page itself as it belongs to SV.

website/docs/r/data_forwarding_rule.html.markdown
namangoya: Updated the docs in the [latest](https://github.com/SumoLogic/terraform-provider-sumologic/pull/691/commits/651ae5560867ff1dc5318edd273fe2ae42588b39) commit.

website/docs/r/data_forwarding_rule.html.markdown
namangoya: Updated the docs in the [latest](https://github.com/SumoLogic/terraform-provider-sumologic/pull/691/commits/651ae5560867ff1dc5318edd273fe2ae42588b39) commit.

website/docs/r/scheduled_view.html.markdown
```diff
@@ -44,6 +45,7 @@ The following arguments are supported:
 The following attributes are exported:
 
 - `id` - The internal ID of the scheduled view.
+- `index_id` - The Index ID of the scheduled view. It remains unchanged during resource updates, and any manual modifications will be disregarded. While it’s not mandatory, we recommend to ignore this via `ignore_changes = [index_id]`.
```
sumovishal, line 48: Now we have made this field as computed you can drop the "modification" and ignore_changes part

website/docs/r/data_forwarding_rule.html.markdown
```diff
@@ -9,25 +9,61 @@ description: |-
...
+For Scheduled Views
+```hcl
+resource "sumologic_scheduled_view" "failed_connections" {
+  index_name = "failed_connections"
+  query = "_sourceCategory=fire | count"
+  start_time = "2024-09-01T00:00:00Z"
+  retention_period = 1
+  lifecycle {
+    prevent_destroy = true
+    ignore_changes = [index_id]
```
sumovishal, line 41: This should not needed now?

website/docs/r/scheduled_view.html.markdown
namangoya: Yeah I have just kept if for the fact that if someone explicitly tries to set `index_id = "000000ABCDE..." (some random value)` in the resource definition, then with the help of `ignore_changes` that will be ignored.

website/docs/r/data_forwarding_rule.html.markdown
namangoya: Yeah I have just kept if for the fact that if someone explicitly tries to set index_id = "000000ABCDE..." (some random value) in the resource definition, then with the help of ignore_changes that will be ignored.

sjain05 (approved):

PR #692, opened 2024-09-20
Fix analytics_tier case sensitivity in resource_sumologic_partition

ssharma-sumologic:
For flex tier, non-lowercase string was causing issue in setting the `isIncludedInDefaultSearch` . Fixed that now.
Also added a diff suppress function for analytics tier to prevent diffs due to case insensitivity.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## 3.0.0 (Unreleased)
 DEPRECATIONS:
 * resource_sumologic_ingest_budget : Deprecated in favour of `resource_sumologic_ingest_budget_v2`.
+
+## 2.31.5 (Unreleased)
 BUG FIXES:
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## 3.0.0 (Unreleased)
 DEPRECATIONS:
 * resource_sumologic_ingest_budget : Deprecated in favour of `resource_sumologic_ingest_budget_v2`.
+BUG FIXES:
```
ashpaliwal, line 4: This is bug fix and we can be released as part of 2.31.5

ssharma-sumologic: done

sumo-mastermanas (approved):
LGTM

sjain05 (approved):

PR #693, opened 2024-09-24
Remove regex check in name field in log search resource

ssharma-sumologic:
The regex in log search name was outdated, we are using a regex allowing names with more broader fields with api. With this change we are removing regex validation in name field in terraform, since we want the api to be the single source of truth. 
The api already has regex pattern checks so the change is safe, now instead of failing at plan generation, terraform will fail at apply stage if name is invalid as per regex. 

Tested the changes by creating resource.


sumo-mastermanas (approved):
LGTM

PR #709, opened 2025-01-03
SUMO-252740 Add method to update status of fields resource

priyanshgoel:
In the APIs exposed by Sumologic fields, we have two APIs /enable and /disable to enable or disable the state of the fields.  So, in this PR, we have build a method to update only the state of the field. Other variables are still not updatable. 

Wrote test for update and tested it via acceptance tests.

### DIFF
--- a/sumologic/resource_sumologic_field.go
+++ b/sumologic/resource_sumologic_field.go
@@ -137,10 +137,6 @@ func resourceSumologicFieldUpdate(d *schema.ResourceData, meta interface{}) erro
 		return err
 	}
 
-	if f.FieldName != name || f.DataType != tpe {
-		return errors.New("Only state field is updatable")
-	}
-
 	if state == "Enabled" {
 		err := c.EnableField(id)
 		if err != nil {
--- a/sumologic/resource_sumologic_field.go
+++ b/sumologic/resource_sumologic_field.go
@@ -122,7 +122,6 @@ func resourceSumologicFieldUpdate(d *schema.ResourceData, meta interface{}) erro
 
 	id := d.Get("field_id").(string)
 	name := d.Get("field_name").(string)
-	tpe := d.Get("data_type").(string)
 	state := d.Get("state").(string)
 	if id == "" {
 		newId, err := c.FindFieldId(name)
@@ -131,7 +130,7 @@ func resourceSumologicFieldUpdate(d *schema.ResourceData, meta interface{}) erro
 		}
 		id = newId
 	}
-	f, err := c.GetField(id)
+	_, err := c.GetField(id)
 
 	if err != nil {
 		return err
--- a/sumologic/resource_sumologic_field.go
+++ b/sumologic/resource_sumologic_field.go
@@ -10,9 +10,8 @@ func resourceSumologicField() *schema.Resource {
 	return &schema.Resource{
 		Create: resourceSumologicFieldCreate,
 		Read:   resourceSumologicFieldRead,
-		Update: resourceSumologicFieldUpdate 
+		Update: resourceSumologicFieldUpdate,
 		Delete: resourceSumologicFieldDelete,
-		Update: 
 		Importer: &schema.ResourceImporter{
 			State: resourceSumologicFieldImport,
 		},
@@ -132,7 +131,7 @@ func resourceSumologicFieldUpdate(d *schema.ResourceData, meta interface{}) erro
 		}
 		id = newId
 	}
-	f,err := c.GetField(id)
+	f, err := c.GetField(id)
 
 	if err != nil {
 		return err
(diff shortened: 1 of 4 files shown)


sumologic/sumologic_field.go
```diff
@@ -94,6 +94,30 @@ func (s *Client) FindFieldId(name string) (string, error) {
...
+	urlWithoutParams := "v1/fields/%s/enable"
+	sprintfArgs := []interface{}{}
+	sprintfArgs = append(sprintfArgs, id)
+
+	urlWithParams := fmt.Sprintf(urlWithoutParams, sprintfArgs...)
+
+	_, err := s.Put(urlWithParams, nil)
+
+	return err
+}
```
raghavxk, line 119: This can be simplified 
```suggestion
func (s *Client) DisableField(id string) error {
	urlWithParams := fmt.Sprintf("v1/fields/%s/disable", id)

	_, err := s.Delete(urlWithParams)
	return err
}

func (s *Client) EnableField(id string) error {
	urlWithParams := fmt.Sprintf("v1/fields/%s/enable", id)

	_, err := s.Put(urlWithParams, nil)
	return err
}

```

raghavxk (approved):

sumologic/resource_sumologic_field.go
```diff
@@ -116,6 +117,39 @@ func resourceSumologicFieldImport(d *schema.ResourceData, meta interface{}) ([]*
 	return []*schema.ResourceData{d}, nil
 }
 
+func resourceSumologicFieldUpdate(d *schema.ResourceData, meta interface{}) error {
+	c := meta.(*Client)
+
+	id := d.Get("field_id").(string)
+	name := d.Get("field_name").(string)
+	tpe := d.Get("data_type").(string)
+	status := d.Get("state").(string)
```
sumovishal, line 126: nit: call it `state` to be consistent with field name?

sumovishal (commented):
Left some comments. Please also mention the testing done to verify this PR.

I will let @ashpaliwal approve the PR given he is the new code owner.

sumovishal:
@ashpaliwal @ErikAtSumo can you update the codeowners file given you folks are taking over the baton? :)

sumologic/resource_sumologic_field.go
```diff
@@ -116,6 +117,39 @@ func resourceSumologicFieldImport(d *schema.ResourceData, meta interface{}) ([]*
 	return []*schema.ResourceData{d}, nil
 }
 
+func resourceSumologicFieldUpdate(d *schema.ResourceData, meta interface{}) error {
```
sumovishal, line 120: Can we add test for update path? We also need to update CHANGELOG.

sumologic/resource_sumologic_field.go
```diff
@@ -116,6 +117,39 @@ func resourceSumologicFieldImport(d *schema.ResourceData, meta interface{}) ([]*
...
+		return errors.New("Only state field is updatable")
+	}
+
+	if status == "Enabled" {
+		return c.EnableField(id)
+	} else if status == "Disabled" {
+		return c.DisableField(id)
+	} else {
+		return errors.New("Invalid value of state field. Only Enabled or Disabled values are accepted")
+	}
```
sumovishal, line 150: You forgot to update the tf state after updating the backend.

sumologic/sumologic_field.go
```diff
@@ -94,6 +94,21 @@ func (s *Client) FindFieldId(name string) (string, error) {
 	return "", fmt.Errorf("Field \"%s\" not found", name)
 }
 
+func (s *Client) DisableField(id string) error {
+	urlWithParams := fmt.Sprintf("v1/fields/%s/disable", id)
+
+	_, err := s.Delete(urlWithParams)
```
sumovishal, line 100: should be `s.Put` instead of delete?

priyanshgoel:
plox test

sumologic/resource_sumologic_field.go
priyanshgoel: Fixed.

sumologic/resource_sumologic_field.go
priyanshgoel: Fixed

sumologic/resource_sumologic_field.go
priyanshgoel: Fixed

sumologic/resource_sumologic_field.go
```diff
@@ -116,6 +117,42 @@ func resourceSumologicFieldImport(d *schema.ResourceData, meta interface{}) ([]*
...
+		}
+		id = newId
+	}
+	f, err := c.GetField(id)
+
+	if err != nil {
+		return err
+	}
+
+	if f.FieldName != name && f.DataType != tpe {
```
ErikAtSumo, line 140: Should this be `||` ?

ErikAtSumo: Also, would you add a test to verify this behavior?

priyanshgoel: yes, thank you for pointing this out.

priyanshgoel: Actually, in process of writing the tests, I figured out that this line/validation is not needed , because we already have forceNew = true for Name and DataType in the Resource Schema. So, probably we can never update and it will always recreate the resource.  @ErikAtSumo  let me know if my understanding is incorrect.

ErikAtSumo: @priyanshgoel I am pretty new to terraform, so I'm reading through some documentation to see what I can find. I think you are correct:

- Suppose the user modifies their Terraform configuration, to set state=Enabled and field_name=SomethingNew
- user runs `terraform apply`
- Then, that should trigger a Create, not an Update, so this line of code should never be reached

priyanshgoel: I have removed this line now.

ErikAtSumo (approved):

PR #717, opened 2025-01-17
Update version to 3.0.1

ErikAtSumo:


ashpaliwal (approved):

sjain05 (approved):
lgtm, but we have now started using `fields` in private app tf provider also so we need to backport this https://github.com/SumoLogic/terraform-provider-sumologic/pull/709 to private provider

ErikAtSumo:
> lgtm, but we have now started using `fields` in private app tf provider also so we need to backport this #709 to private provider

Thanks @sjain05 . I'll make sure there's an internal issue for this.

PR #721, opened 2025-01-29
Release 3.0.2

ErikAtSumo:


CHANGELOG.md
```diff
@@ -1,28 +1,31 @@
 ## X.Y.Z (Unreleased)
 * Add new change notes here
+
+## 3.0.2 (January 29, 2025)
 FEATURES:
 * **New Resource:** sumologic_azure_metrics_source (GH-710)
```
ashpaliwal, line 6: Suggestion - We should keep format in sync, like not have PR in here or have everywhere. IMHO, we don't need PR number

ErikAtSumo: @ashpaliwal "We should keep format in sync" I agree. It looks like the (3.0.0, 3.0.1) formatting is different from all of the past formatting, so I synchronized them with the established pattern. Ideally, these "Release" PRs will only add lines 3 and 4, shown above.

ashpaliwal: My comment is specific to Line 6 only, rest is ok.

ErikAtSumo: It's also there for every line pre-3.0.0, so I'm inclined to allow them to remain, but I agree, it seems unnecessary. \

ashpaliwal (approved):
lgtm

PR #722, opened 2025-01-31
Update version to 3.0.3

ErikAtSumo:


ashpaliwal:
@ErikAtSumo let's just double check it's ready for GA

ashpaliwal (approved):
lgtm, once we confirm it's ready for GA, merge it. removing the bottleneck

ErikAtSumo:
@sukhhanda1 Is there any reason that the Metrics Search v2 should not be released?

sukhhanda1:
> @sukhhanda1 Is there any reason that the Metrics Search v2 should not be released?

No there isn't. We can release it right away.

PR #730, opened 2025-02-27
Fix the use_versioned_api noise

ErikAtSumo:
For polling sources that do not use an S3 bucket, the `use_versioned_api` path setting means nothing. Unfortunately, they still generate some noise in the `terraform plan` output, because terraform thinks the missing values ought to be set to something.

By setting the use_versioned_api default value to `nil`, terraform is able to distinguish between the case of "use_versioned_api is false" and "use_versioned_api is not set at all". This should remove the noise.

Several acceptance tests were configured to tolerate this noise with `ExpectNonEmptyPlan: true`. That tolerance is also removed in this PR.

Manual tests:

- create a sumologic_aws_xray_source
   - in .tf, path has no use_versioned_api
   - before: created resource has use_versioned_api = false
   - after: created resource has no use_versioned_api
- create a sumologic_cloudtrail_source
   - in .tf, path has no use_versioned_api
   - before: created resource has use_versioned_api = false
   - after: created resource has no use_versioned_api
- sumologic_cloudtrail_source
   - in .tf, path has use_versioned_api = false
   - before: created resource has use_versioned_api = false
   - after: created resource has use_versioned_api = false

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 ## X.Y.Z (Unreleased)
 * Add new change notes here
 
+BUG FIXES:
+* Polling source paths without `use_versioned_api` will not show an incorrect diff in the `terraform plan`
+
 ## 3.0.4 (February 4, 2025)
 BUG FIXES:
 * Fixed issue with release artifacts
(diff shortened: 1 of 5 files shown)


vsinghal13 (approved):
Thanks! I have also verified by creating test sources locally on my end.

PR #745, opened 2025-03-25
SUMO-256375: Update webhook_search_notification logic to send null payload instead of empty string when missing

ShahzaibA123:


sumologic/sumologic_log_search.go
```diff
@@ -157,11 +157,11 @@ type ServiceNowSearchNotification struct {
...
 
 type WebhookSearchNotification struct {
-	TaskType          string `json:"taskType"`
-	WebhookId         string `json:"webhookId"`
-	Payload           string `json:"payload"`
-	ItemizeAlerts     bool   `json:"itemizeAlerts"`
-	MaxItemizedAlerts int    `json:"maxItemizedAlerts"`
+	TaskType          string      `json:"taskType"`
+	WebhookId         string      `json:"webhookId"`
+	Payload           interface{} `json:"payload"`
```
ErikAtSumo, line 162: @ShahzaibA123 How would you feel about using `*string` instead, for some extra type safety?

```
    Payload           *string `json:"payload"`
```

```
    var payload *string
    if p, ok := tfWebhookSearchNotification["payload"].(string); ok && p != "" {
        payload = &p
    }
```

ShahzaibA123: Good catch! Go isn't a language I've used but yeah makes sense to use a pointer. Went ahead and made the change and test looks good.

ErikAtSumo (approved):

PR #747, opened 2025-03-31
Preserve assigned users when updating RoleV2

ErikAtSumo:
RoleV2 is missing some special logic for preserving the list of users assigned to that role, when you update some other properties of the role. This PR fixes the problem by copying some logic from the v1 Role:

https://github.com/SumoLogic/terraform-provider-sumologic/blob/6ca2ccfc8fcb743f77f23d6402038ccf00b3b26a/sumologic/resource_sumologic_role.go#L108-L110

Users can be assigned a role by adding the role ID to a `sumologic_user` -> `role_ids`.


adisuj (approved):
lgtm

PR #748, opened 2025-04-01
Sumo 257823 remove realtime reference

rmamilla-sumo:
Removed reference of `Realtime` option from Scheduled search TF documentation page as it is deprecated right now.


ErikAtSumo:
Thanks @rmamilla-sumo . I see that `"RealTime"` is still accepted by the `ValidationFunc` for `schedule_type`. When will it be safe to completely remove it?

rmamilla-sumo:
> Thanks @rmamilla-sumo . I see that `"RealTime"` is still accepted by the `ValidationFunc` for `schedule_type`. When will it be safe to completely remove it?

Yes, we have to keep this still because there are still exceptions made for some cx's to allow real time scheduled searches. Removing this would break their functionality.

PR #750, opened 2025-04-08
Improve error message when an API URL is constructed with missing parameters

ErikAtSumo:


sjain05 (approved):
I don't understand the nitty-gritty but 
1. compared current logic with before
2. trust our test as we have a lot of those

PR #757, opened 2025-04-23
SUMO-258848: disallow updates to partition's analytics tier other than in case of flex migration

raghavxk:
This PR addresses an issue with how the `analytics_tier` field is handled in the partition resource.

The `analytics_tier` is intended to be set only at the time of partition creation and cannot be modified afterward by the user. However, the current behavior allows Terraform to issue an update when the analytics_tier value is changed in the configuration. This leads to an inconsistent state where the update is accepted, even though the operation is invalid and has no effect in the backend.

That said, there is one valid scenario where the analytics_tier may change implicitly: when an account is migrated to Flex. In such cases, we should allow users to update their Terraform configuration to reflect the actual state in Sumo Logic.

This change ensures that:

- User-initiated updates to the analytics_tier are blocked if the value differs from what's currently in Sumo Logic.
- Terraform can still reconcile the state when the backend value changes outside of user control (e.g., due to Flex migration).


sumologic/resource_sumologic_partition.go
```diff
@@ -183,3 +196,15 @@ func resourceToPartition(d *schema.ResourceData) Partition {
...
+func areAnalyticsTierEqual(a, b *string) bool {
+	if a == nil && b == nil {
+		return true
+	}
+
+	if a == nil || b == nil {
+		return false
+	}
+
+	return strings.EqualFold(*a, *b)
```
stripathi-sumo, line 209:
```suggestion
	return (a == nil && b == nil) || (a != nil && b != nil && strings.EqualFold(*a, *b))
```

sumologic/resource_sumologic_partition.go
```diff
@@ -135,11 +136,23 @@ func resourceSumologicPartitionDelete(d *schema.ResourceData, meta interface{})
...
 func resourceSumologicPartitionUpdate(d *schema.ResourceData, meta interface{}) error {
+	c := meta.(*Client)
+
+	partitionId := d.Id()
+
 	spartition := resourceToPartition(d)
+	if d.HasChange("analytics_tier") {
+		currPartitionState, err := c.GetPartition(partitionId)
+		if err != nil {
+			return fmt.Errorf("error loading partition with id %s for analytics_tier update validation", partitionId)
```
stripathi-sumo, line 147: add a UT for this as well.

ErikAtSumo:
Thanks for this fix. The logic looks good to me. Would you update the change log and the html doc page?

sumologic/resource_sumologic_partition.go
stripathi-sumo: this is just a shorter way to write the same condition. IMO reduce congnitive complexity and is more terse.

ErikAtSumo: Interesting, I feel like it increases the cognitive complexity 😆 

Either way, what do you think about changing `a == nil && b == nil` with `a == b`? This would cover the case where they're both nil, AND the case where they both point to the same string object.

stripathi-sumo: > Either way, what do you think about changing `a == nil && b == nil` with `a == b`? This would cover the case where they're both nil, AND the case where they both point to the same string object.

Yes that is better.

raghavxk: On the cognitive load thing, I agree with Erik - what Sathyam is suggesting seems difficult to read to me too, in comparison. 

I can add `a == b` for case where they both point to same object but isn't that already covered with `strings.EqualFold(*a, *b)`?

ErikAtSumo: `a == b` is both a short-circuit, and a reduction in mental load. Before:

```
if a == nil && b == nil {
  return true
}
// or
return (a == nil && b == nil) || ...rest...
```

after:

```
if a == b {
  return true
}
// or
return (a == b) || ...rest...
```

sumologic/resource_sumologic_partition.go
raghavxk: I don't think that is possible with current testing framework. Adding support for it, seems like beyond the scope of change here.

stripathi-sumo: then how come we have a test for `// Update analytics tier to a different value and assert error` ?

sumologic/resource_sumologic_partition.go
stripathi-sumo: isn't there a scala like `Option` available in go?

raghavxk: No

sumologic/resource_sumologic_partition.go
raghavxk: TF test utils allow running test like real world scenario. I can't mock an API failure.

raghavxk:
> Thanks for this fix. The logic looks good to me. Would you update the change log and the html doc page?

I'll be adding these. I wanted to wait till change was finalised.

sumologic/resource_sumologic_partition.go
stripathi-sumo: the reason why I said cognitive complexity is high because by the time second condition `a == nil || b == nil` is checked the first is false. so in reality the expanded boolean check is `!(a == nil && b == nil) && a == nil || b == nil` which comes out to `(a == nil && b != nil) || (a != nil && b == nil)`. that is why I thought its simpler to have that in one line.

stripathi-sumo (approved):

ErikAtSumo:
One more file before I can approve:
https://github.com/SumoLogic/terraform-provider-sumologic/blob/master/website/docs/r/partition.html.markdown
Would you explain the analytics_tier behavior in this doc?

ErikAtSumo (approved):
Perfect. Thanks for the fix!

PR #758, opened 2025-04-24
ignore running test for changes to static content

raghavxk:
We can avoid running TF acceptance tests in case only static contents have been changed.


ErikAtSumo (approved):
Good improvement 👍

PR #636, opened 2024-03-22
feat: s3 data forwarding

imranismail:
Support use-case to deploy S3 data forwarding destination and rules

```terraform
resource "aws_s3_bucket" "this" {
  bucket = "${terraform.workspace}-sumologic"
}

resource "aws_s3_bucket_public_access_block" "this" {
  bucket                  = aws_s3_bucket.this.id
  block_public_acls       = true
  block_public_policy     = true
  ignore_public_acls      = true
  restrict_public_buckets = true
}

data "aws_iam_policy_document" "trust" {
  statement {
    actions = ["sts:AssumeRole"]
    effect  = "Allow"

    principals {
      type        = "AWS"
      identifiers = ["arn:aws:iam::926226587429:root"]
    }

    condition {
      test     = "StringEquals"
      variable = "sts:ExternalId"
      values   = ["au:0000000000595DED"]
    }
  }
}

resource "aws_iam_role" "this" {
  name               = "${terraform.workspace}-sumologic-s3-writer"
  assume_role_policy = data.aws_iam_policy_document.trust.json
}

data "aws_iam_policy_document" "role" {
  statement {
    actions   = ["s3:PutObject"]
    resources = ["${aws_s3_bucket.this.arn}/*"]
  }
}

resource "aws_iam_role_policy" "this" {
  name   = "sumologic-s3-writer"
  role   = aws_iam_role.this.id
  policy = data.aws_iam_policy_document.role.json
}

data "sumologic_partitions" "this" {}

resource "sumologic_s3_data_forwarding_destination" "this" {
  depends_on          = [aws_iam_role_policy.this]
  name                = "${terraform.workspace}-sumologic"
  bucket_name         = aws_s3_bucket.this.bucket
  authentication_mode = "RoleBased"
  role_arn            = aws_iam_role.this.arn
  encrypted           = true
}


locals {
  partitions = {
    for partition in data.sumologic_partitions.this.partitions : partition.name => partition
    if startswith(partition.name, "${terraform.workspace}_eks_") && partition.analytics_tier != "infrequent"
  }
}


resource "sumologic_s3_data_forwarding_rule" "this" {
  for_each       = local.partitions
  index_id       = each.value.id
  destination_id = sumologic_s3_data_forwarding_destination.this.id
  file_format    = "{index}/{day}/{hour}_{minute}_{second}"
}

```

PR #762, opened 2025-04-30
Erikatsumo datasource partition

ErikAtSumo:
This stale PR introduces two new resources, and two new data sources:

https://github.com/SumoLogic/terraform-provider-sumologic/pull/636

The resources have already been merged last year. This PR adds the data sources. Once this PR is merged, the stale one can be safely closed.


sumologic/data_source_sumologic_partition.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+}
+
+func dataSourceSumologicPartitionRead(d *schema.ResourceData, meta interface{}) error {
+	c := meta.(*Client)
+
+	var err error
+	var spartition *Partition
+
+	if rid, ok := d.GetOk("id"); ok {
```
raghavxk, line 75: nit: we can reduce nesting if we check for empty id and do an early return with an error.

sumologic/data_source_sumologic_partition_test.go
```diff
@@ -0,0 +1,42 @@
...
+				Config: testDataSourceSumologicPartitionConfig,
+				Check: resource.ComposeTestCheckFunc(
+					testAccDataSourcePartitionCheck("data.sumologic_partition.test"),
+				),
+			},
+		},
+	})
+}
+
+func testAccDataSourcePartitionCheck(name string) resource.TestCheckFunc {
```
raghavxk, line 24: Any reason why we are not asserting if all attributes on schema are set?

sumologic/data_source_sumologic_partitions.go
```diff
@@ -0,0 +1,58 @@
...
+			"is_active":                     spartition.IsActive,
+			"index_type":                    spartition.IndexType,
+			"is_included_in_default_search": spartition.IsIncludedInDefaultSearch,
+		}
+
+		partitions = append(partitions, partition)
+	}
+
+	d.Set("partitions", partitions)
+	d.SetId(c.BaseURL.Host)
```
raghavxk, line 55: I am not sure about how we handle ids on a data source but BaseURL.Host can be common across mulltiple resources. Will that cause any issues? 
This is more of an question for my understanding than a review comment.

sumologic/sumologic_partition.go
```diff
@@ -77,3 +118,13 @@ type Partition struct {
...
+
+type ListPartitionResp struct {
+	Data []Partition `json:"data"`
+	Next string      `json:"next"`
+}
+
+func (s *ListPartitionResp) Reset() {
+	s.Data = nil
+	s.Next = ""
+}
```
raghavxk, line 130: Given, these structs have no purpose outside of API call method, should we scope em to function level? 
Since, all code is within a single package, we can't use package level scope here.

sumologic/data_source_sumologic_partitions.go
ErikAtSumo: Good catch. This is indeed a mistake.

sumologic/sumologic_partition.go
ErikAtSumo: We could, but I think it would bloat the function without really adding much value. Is there a risk you're trying to protect against? Perhaps reduce the clutter in an IDE's autocompletion?

sumologic/data_source_sumologic_partition.go
```diff
@@ -0,0 +1,102 @@
...
+		id := rid.(string)
+		spartition, err = c.GetPartition(id)
+		if err != nil {
```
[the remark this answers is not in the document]
ErikAtSumo: Changed to:
`return fmt.Errorf("error retrieving partition with id %v: %v", id, err)`

sumologic/data_source_sumologic_partition.go
ErikAtSumo: Fixed

sumologic/data_source_sumologic_partitions.go
ErikAtSumo: Fixed.

sumologic/data_source_sumologic_partition_test.go
raghavxk: Yes, I think we should do that. We can connect if you need help with this.

sumologic/sumologic_partition.go
raghavxk: > Is there a risk you're trying to protect against? Perhaps reduce the clutter in an IDE's autocompletion?

Nothing major. I just wanted to keep API surface minimal but this repo isn't going to be imported anywhere so shouldn't matter.

sumologic/data_source_sumologic_partition.go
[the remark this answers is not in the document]
raghavxk: Thank you

sumologic/data_source_sumologic_partition_test.go
ErikAtSumo: @raghavxk I found a different partition in the test environment which does have a routing expression, so I updated the test to use that partition, and also verify the actual values of all the properties, instead of just verifying that they are set. How does this look?

raghavxk: lgtm!

PR #763, opened 2025-05-01
Update Match Rule Documentation: Fixed Indentation for `mapping` fields

anishrajan25:
## Issue
In the Terraform Provider documentation for [SIEM Match Rule](https://registry.terraform.io/providers/SumoLogic/sumologic/latest/docs/resources/cse_match_rule), the fields `type`, `from`, and `to` are listed at the same level as their parent, [mapping](https://registry.terraform.io/providers/SumoLogic/sumologic/latest/docs/resources/cse_match_rule#mapping-4) field.

## Fix
Updated the documentation for Match Rule to have proper identation for the above mentioned fields.


ErikAtSumo:
Verified on https://registry.terraform.io/tools/doc-preview

PR #764, opened 2025-05-02
Disallow changing a user's email address

### LINKED ISSUE #671: Changing a user's email address doesn't force a resource recreation and falsely succeeds
I changed a user from

```hcl
resource "sumologic_user" "jdoe" {
  first_name  = "John"
  last_name   = "Doe"
  email       = "john.doet@example.com"
  transfer_to = ""
  role_ids    = [sumologic_role.example-production.id]
  is_active   = true
}
```

to

```hcl
resource "sumologic_user" "jdoe" {
  first_name  = "John"
  last_name   = "Doe"
  email       = "john.doe@example.com"
  transfer_to = ""
  role_ids    = [sumologic_role.example-production.id]
  is_active   = true
}
```

The provider showed that as an in-place change, Terraform claimed to apply it successfully, but then the change was not made. As I can see in the web console's user settings, email addresses cannot be changed on existing users. Attempting to change a user's email address should either delete and recreate the user, or it should fail since the email address was unchanged.

ErikAtSumo:
The API for updating a user does not allow the email address to be modified. If someone tries to modify the email of a sumologic_user, then the provider will silently "pass", without making the desired change.

The new behavior informs the user that the email address may not be changed.

Closes #671

Note that #671 requests `ForceNew: true` for a user's email. I discussed with Kevin, and we agree that recreating a user unintentionally can have adverse consequences, so it's safer to just stop the operation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,14 +1,17 @@
 ## X.Y.Z (Unreleased)
 * Add new change notes here
 
+BUG FIXES:
+* When the email of a sumologic_user is changed, instead of silently doing nothing, an error will be raised to the user
+
+DOCS:
+* Updated content docs to include examples of different alerts
+
 ## 3.0.10 (May 6, 2025)
 
 BUG FIXES:
 * Fix the log mapping resource to no longer error out when empty strings are included in skipped_values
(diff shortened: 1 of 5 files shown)


sjain05 (approved):
test is failing, don't merge without test passing. I have correct the repo to only run tests when required

PR #765, opened 2025-05-06
SUMO-259610: Add other examples of using the content resource

ShahzaibA123:

PR #769, opened 2025-05-09
Add other examples of using the content resource

sjain05:
Changes copied from https://github.com/SumoLogic/terraform-provider-sumologic/pull/765 due to git checks issues


ShahzaibA123 (approved):
LGTM

PR #772, opened 2025-05-20
Update version to 3.0.11

ErikAtSumo:


sjain05 (commented):
lgtm, the reason this was not working is that the deleted code was wrong in the first place

PR #781, opened 2025-06-30
Feat: Add timezone support for scheduled view

ssharma-sumologic:
Added support for timezone in scheduled view, now user's can specify the timezone while creating scheduled view, and update it as well.
Wrote acceptance tests for scheduled views 

Testing performed : Tested locally on environment, ran acceptance tests locally


AbhishekSumo (approved):
LGTM as well

PR #783, opened 2025-07-14
add local windows event log source

### LINKED ISSUE #581: local windows event log source
Recently a local file source was added, but the provider does not have support for local windows event log collection which is important for customers using terraform to configure installed collectors on Windows.

Please add local event source similar to this but for windows event log:
https://github.com/SumoLogic/terraform-provider-sumologic/blob/master/sumologic/resource_sumologic_local_file_source.go 
https://github.com/SumoLogic/terraform-provider-sumologic/blob/master/sumologic/sumologic_local_file_source.go 

Required fields are defined here over the common shared ones such as name, category etc  https://help.sumologic.com/docs/send-data/use-json-configure-sources/json-parameters-installed-sources/#local-windows-event-logsource

rjury-sumo:
This PR adds support for a local windows events source for the installed collector. Like the local file source this is a special case as it can only be applied to existing installed collectors rather than to a hosted collector.

I don't have tests for this (the local file source doesn't either) - but have successfully been able to create or modify sources on a windows instance with installed collector installed using the two example configurations in the mardkdown file.

This was requested by one sumo customer and has an internal JIRA field request https://sumologic.atlassian.net/browse/FIELD-95

Closes #581

### DIFF
--- a/sumologic/resource_sumologic_local_windows_event_log_source.go
+++ b/sumologic/resource_sumologic_local_windows_event_log_source.go
@@ -1,0 +1,151 @@
+package sumologic
+
+import (
+	"fmt"
+	"log"
+	"strconv"
+
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
+	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
+)
+
+func resourceSumologicLocalWindowsEventLogSource() *schema.Resource {
+	LocalWindowsEventLogSource := resourceSumologicSource()
+	LocalWindowsEventLogSource.Create = resourceSumologicLocalWindowsEventLogSourceCreate
+	LocalWindowsEventLogSource.Read = resourceSumologicLocalWindowsEventLogSourceRead
+	LocalWindowsEventLogSource.Update = resourceSumologicLocalWindowsEventLogSourceUpdate
+	// LocalWindowsEventLogSource.Delete: resourceSumologicSourceDelete
+	LocalWindowsEventLogSource.Importer = &schema.ResourceImporter{
+		State: resourceSumologicSourceImport,
+	}
+
+	// Windows Event Log specific fields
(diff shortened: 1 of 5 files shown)


sumologic/resource_sumologic_local_windows_event_log_source.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	LocalWindowsEventLogSource.Schema["deny_list"] = &schema.Schema{
+		Type:        schema.TypeString,
+		Optional:    true,
+		Elem:        &schema.Schema{Type: schema.TypeString},
+		Description: "Comma-separated list of event IDs to deny",
+	}
+
+	// "allowlist":  is not implmemented yet
```
vsinghal13, line 59: can we get the `allow_list` also added?

vsinghal13 (commented):
thanks @rjury-sumo for creating the PR. LGTM except one point

sumologic/resource_sumologic_local_windows_event_log_source.go
rjury-sumo: I've done that, tidied up a few loose ends and fixed the two linting bugs

rjury-sumo:
If you do want to test it works fine with this tf if you have an installed collector 

`terraform {
  required_providers {
    sumologic = {
      source = "sumologic.com/dev/sumologic"
      version = "~> 1.0.0"
    }
  }
}

data "sumologic_collector" "installed_collector" {
  name = "terraform_source_testing"
}

resource "sumologic_local_windows_event_log_source" "local" {
  name             = "windows_logs"
  description      = "windows system and application logs in json format"
  category         = "/os/windows/events"
  collector_id     = "${data.sumologic_collector.installed_collector.id}"
  log_names  = ["System","Application","Microsoft-Windows-PowerShell/Operational", "Microsoft-Windows-TaskScheduler/Operational"]
  deny_list = "9999,7890"
 // allow_list = "1000,2000"
  event_format = 1 // 0 = XML, 1 = JSON
}
`

PR #789, opened 2025-07-17
Define possible values for `value_type` in `field` schema

### LINKED ISSUE #788: Log mapping schema for `field` is unclear
The field `value_type` within the `field` schema would benefit greatly with additional information. This is an important field to determine what kind of log mapping is needed. Documentation here is especially critical because the possible values are not intuitively named. The prime example of this is if you want a "structured" log mapping for a particular field, you must set `value_type` to equal 'sumofield'. 

Additionally, it could be made more clear that setting this field to null is how to configure a "standard" log mapping for a field.

darianm23:
Resolves #788 

This field should be well defined with possible values given that some of the value names are not intuitive.

### DIFF
--- a/website/docs/r/cse_log_mapping.html.markdown
+++ b/website/docs/r/cse_log_mapping.html.markdown
@@ -63,7 +63,7 @@ The following arguments are supported:
 
 ### Schema for `field`
 - `name` - (Required) Name of the field.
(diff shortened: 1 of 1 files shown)


ErikAtSumo:
Thanks @darianm23 . Are these values also in the Sumo Logic API documentation? I'm having trouble finding it.

PR #792, opened 2025-07-23
DET-727: Added rule tuning expression ids field

antonymartinsumo:
DET-727: Added support for rule tuning expression ids for all rule types


aj-sumo:
Please add an entry to CHANGELOG.md

PR #793, opened 2025-07-23
unmarshal computed property `field_names` on `field_extraction_rule`

hashi-breed:
The REST API documentation indicates that the list of identifiers projected from the parse
expression configured in the `field_extraction_rule` resource. The json data for
`fieldNames` will contain an array of strings for each named field projection. 

we added `FieldNames` to the the terraform schema as `field_names`, 
extended the client, updated the acc tests, and updated docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## X.Y.Z (Unreleased)
-* Add new change notes here
+* return computed `field_names` property for resource `field_extraction_rule`
 
 FEATURES:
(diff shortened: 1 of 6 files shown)


ErikAtSumo:
@SumoLogic/tf-platform-team I've been chatting with Ryan on Slack. To run the acceptance tests, I duplicated this PR and fixed the change log here: https://github.com/SumoLogic/terraform-provider-sumologic/pull/797. Recommend closing this one after the other is approved and merged.

PR #798, opened 2025-07-31
SUMO-266317: Fix `selectionType` handling for Role V2 to avoid `MatchError` 

AbhishekSumo:
### Issue

- The Sumo Logic Roles V2 API omits `selectionType` for default roles.
- In the provider `Read` function:
  - When the API omits the field, `json.Unmarshal` assigns the Go zero value (`""`) to `SelectionType`.
  - The provider calls `d.Set("selection_type", roleV2.SelectionType)` with that empty string.
  - The Terraform SDK interprets empty string for an optional `TypeString` as **unset** and stores `null` in state.
- On the next `terraform apply`:
  - `d.Get("selection_type")` returns `""` (SDK converts `null → ""`).
  - The provider sends `""` to the backend.
  - The backend’s `selectionTypeToAvroSelectionType` throws a `scala.MatchError` for empty string.

### Root Cause

- `omitempty` only affects JSON **encoding**, not decoding.
- When the API omits `selectionType`, Go’s `json.Unmarshal` sets it to the zero value `""`.
- Terraform interprets `""` as unset for optional attributes and stores `null`.
- On update, the provider sends `""` to the API instead of a valid value.

### Fix

1. **Normalize in `Read`:**
 - When `SelectionType == ""`, default it to `"All"` before `d.Set`.
 - Prevents Terraform from storing `null` and avoids perpetual diffs.

2. **Force valid value in `Update`:**
 - Before calling the API, if `SelectionType == ""`, set it to `"All"`.
 - Ensures a valid value is always sent and avoids `MatchError`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## X.Y.Z (Unreleased)
-
+* Add new change notes here
 BUG FIXES:
 * Corrected selection_type configuration in RolesV2 resources.
 


atishya-22 (commented):
Can we add test for this in `resource_sumologic_role_v2_test.go`

sumologic/resource_sumologic_role_v2.go
```diff
@@ -123,6 +127,11 @@ func resourceSumologicRoleV2Update(d *schema.ResourceData, meta interface{}) err
 
 	roleV2 := resourceToRoleV2(d)
 
+	if roleV2.SelectionType == "" {
+       roleV2.SelectionType = "All"
+    }
```
ErikAtSumo, line 132: something isn't right with the formatting. I think this repo uses tabs instead of spaces

sumologic/data_source_sumologic_monitor_folder_test.go
```diff
@@ -8,12 +8,12 @@ import (
 )
 
 func TestAccDataSourceMonitorFolder_basic(t *testing.T) {
-    resource.Test(t, resource.TestCase{
-        PreCheck:  func() { testAccPreCheck(t) },
```
AbhishekSumo, line 12: GoFmt Fixes made changes in this file, should be fine to push these indentation changes.

If Not I can remove these files

sumologic/resource_sumologic_role_v2.go
AbhishekSumo: Fixed now.

PR #801, opened 2025-08-01
SUMO-267529: Update sumologic role v2 to public from beta

dineshmeka701557:


ErikAtSumo:
Yes, please add a line to the changelog. It seems appropriate for the `FEATURES` section - something like "sumologic_role_v2 resource is promoted from beta to general release" or similar

PR #803, opened 2025-08-08
DET-729: Added suppressSpaceDiff for expression

antonymartinsumo:
Added a fix to suppress any diff for whitespaces in the match_expression/expression field


joescii (commented):
Is it possible for this PR to include tests validating this change works?

PR #817, opened 2025-10-15
add auto_pause_enabled support

prajalb:
### Description
adding support for auto_pause_enabled in create/get/update flow for SVs

tested on local env

### DIFF
--- a/sumologic/sumologic_scheduled_view.go
+++ b/sumologic/sumologic_scheduled_view.go
@@ -67,5 +67,5 @@ type ScheduledView struct {
 	ParsingMode                      string    `json:"parsingMode"`
 	ReduceRetentionPeriodImmediately bool      `json:"reduceRetentionPeriodImmediately"`
 	TimeZone                         string    `json:"timeZone,omitempty"`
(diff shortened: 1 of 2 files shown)


ssharma-sumologic (commented):
I believe

sumologic/sumologic_scheduled_view.go
```diff
@@ -67,4 +67,5 @@ type ScheduledView struct {
 	ParsingMode                      string    `json:"parsingMode"`
 	ReduceRetentionPeriodImmediately bool      `json:"reduceRetentionPeriodImmediately"`
 	TimeZone                         string    `json:"timeZone,omitempty"`
+	AutoPauseEnabled                 bool      `json:"autoPauseEnabled,omitempty"`
```
ssharma-sumologic: I believe the omitempty here could have unexpected consequence , iirc the code will treat false as empty and so if user has mentioned autoPauseEnabled as false, then it won't be set correctly.

prajalb: removed omitempty, it was indeed causing issues

PR #866, opened 2026-04-29
CSRE-2371: add region field to sumologic_kinesis_log_source authentication schema

ajsaclayan-sumo:
### What

Add `region` field to the `authentication` block schema of `sumologic_kinesis_log_source`.

### Why

The Go implementation in `getKinesisLogAuthentication()` already handles `auth["region"]` for both `S3BucketAuthentication` and `AWSRoleBasedAuthentication`, but the field was missing from the Terraform schema definition. This made it impossible to specify a region via Terraform configuration, causing `KinesisLogPath` S3 authentication to fail for non-standard AWS partitions such as EU Sovereign Cloud (`eusc-de-east-1`).

Without `region`, Sumo Logic defaults to standard AWS S3 endpoints (`s3.amazonaws.com`), which are unreachable from sovereign cloud partitions.

### Changes

- Added `region` (Optional, TypeString) to `authentication` schema in `resource_sumologic_kinesis_log_source.go`
- Added acceptance test `TestAccSumologicKinesisLogSource_s3BucketAuthWithRegion` covering `S3BucketAuthentication` with `region`
- Updated docs with `region` argument reference and EU Sovereign Cloud example
- Updated CHANGELOG

### Testing

- [x] Compiled the code
- [x] Ran UTs locally
- [ ] Tested on EpD

🤖 Generated with [Claude Code](https://claude.com/claude-code)


website/docs/r/kinesis_log_source.html.markdown
```diff
@@ -34,6 +34,28 @@ __IMPORTANT:__ The AWS credentials are stored in plain-text in the state. This i
     }
   }
 
+  # EU Sovereign Cloud example using access key authentication with explicit region
```
maimaisie: I don't think we need to have a separate example just for ESC because this is a user-facing doc. we can add region in the existing example and use a standard region such as us-east-1

maimaisie (approved):

website/docs/r/kinesis_log_source.html.markdown
ajsaclayan-sumo: Done!

PR #879, opened 2026-06-30
CSIEM-2549:fix items hash to exclude computed id, preventing unnecessary plan diffs

rsha-sumo:
### Description
Problem:

Every terraform apply on a sumologic_cse_match_list with items produced a non-empty plan — deleting and re-creating all items even when nothing changed. This caused unnecessary API churn on every apply.

Root Cause:
 The items field uses schema.TypeSet, which hashes each element to detect changes. The default hash included the id field (which is Computed — empty in config, populated in state after API response). This hash mismatch made Terraform think every item was different on every plan.

Fix:

 Added a custom Set hash function (matchListItemHash) that hashes only user-meaningful fields: value, description, and expiration - excluding the computed id. This makes config items and their state counterparts hash identically when unchanged, producing an empty plan.

### Check list
* [ X] Describe the changes and their purpose
* [X ] Update HTML docs (if applicable)
    * [preview tool](https://registry.terraform.io/tools/doc-preview)
* [ X] Update [`CHANGELOG.md`](../CHANGELOG.md)
* [ X] Check [`CONTRIBUTING.md`](../CONTRIBUTING.md) for anything forgotten

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
-## 3.2.9 (Jun 30, 2026)
+## X.Y.Z (Unreleased)
+* Fixed `sumologic_cse_match_list` producing a non-empty plan on every apply by excluding the computed `id` from the items set
+  hash.
 
+## 3.2.9 (Jun 30, 2026)
 FEATURES:
 * **New Resource:** `sumologic_lambda_invoke_action` - Invoke AWS Lambda functions for enabling S3 logging and auto-enable operations.
 
--- a/sumologic/resource_sumologic_cse_match_list.go
+++ b/sumologic/resource_sumologic_cse_match_list.go
@@ -1,6 +1,7 @@
 package sumologic
 
 import (
+	"bytes"
 	"fmt"
 	"log"
 	"regexp"
@@ -60,6 +61,7 @@ func resourceSumologicCSEMatchList() *schema.Resource {
 			"items": {
 				Type:     schema.TypeSet,
 				Optional: true,
+				Set:      matchListItemHash,
 				Elem: &schema.Resource{
 					Schema: map[string]*schema.Schema{
 						"id": {
@@ -86,6 +88,15 @@ func resourceSumologicCSEMatchList() *schema.Resource {
 			},
 		},
 	}
+}
+
+func matchListItemHash(v interface{}) int {
+	var buf bytes.Buffer
+	m := v.(map[string]interface{})
+	buf.WriteString(m["value"].(string))
+	buf.WriteString(m["description"].(string))
+	buf.WriteString(m["expiration"].(string))
+	return schema.HashString(buf.String())
 }
 
 func resourceSumologicCSEMatchListRead(d *schema.ResourceData, meta interface{}) error {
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
-## 3.2.9 (Jun 30, 2026)
+## X.Y.Z (Unreleased)
```
hreeve-sumo, line 1: Was the X.Y.Z version intentional?

rsha-sumo: I think on release it generates the version

rsha-sumo: but let me double check on this

rsha-sumo: here is the diff on the last commit to master - https://github.com/SumoLogic/terraform-provider-sumologic/commit/a2a1a04932c81821ec3bdc7d777127658429fdb4

rsha-sumo: heres another one https://github.com/SumoLogic/terraform-provider-sumologic/commit/a2a1a04932c81821ec3bdc7d777127658429fdb4

So I think doing this right

hreeve-sumo: Thanks for checking! It looks like they use the X.Y.Z notation until they're ready to cut a release. If this is for a CSI I assume we'll want to release it so the customer can use it, but we can touch base with the owning team about how the release process works after the fact.

hreeve-sumo (approved):

sumologic/resource_sumologic_cse_match_list.go
```diff
@@ -88,6 +90,15 @@ func resourceSumologicCSEMatchList() *schema.Resource {
 	}
 }
 
+func matchListItemHash(v interface{}) int {
+	var buf bytes.Buffer
+	m := v.(map[string]interface{})
```
rquinlivan: This is an unsafe type cast and will panic if the cast fails. You can use the `ok` variable to check if it errors, e.g.:

```go
m, ok := v.(map[string]interface{})
if !ok {
  // print useful error
}
```

rsha-sumo: In practice the SDK always passes a map[string]interface{} here so it won't fail, but defensive coding is reasonable - I will update

CHANGELOG.md
ErikAtSumo: When I set this up a year ago, this was the intent:

1. changelog ALWAYS includes that `## X.Y.Z (Unreleased)` line at the top
2. PRs should add new lines in that section
3. When release is cut, inject the version number and date on a new line, after the `## X.Y.Z (Unreleased)` line, because all of the previously-unreleased changes are now included in a versioned release

rsha-sumo: >When release is cut,

Is this a manual process or automated ?

rsha-sumo: based on these searches looks like it a manual process https://sumologic.slack.com/archives/CAU3N0Y92/p1782841958208829?thread_ts=1781883869.051269&cid=CAU3N0Y92

rmamilla-sumo (approved):

PR #887, opened 2026-08-05
CSIEM-2549: fix perpetual diffs from nil expiration and case-sensitive values

hreeve-sumo:
### Description

Adds some additional logic to the bulk match list update plan diffs that were resulting in an inordinate amount of time spent on plans that yielded no functional differences and could lead to timeouts.

* The hash function now lowercases values before hashing (matching the API's storage behavior) so that case differences won't cause a plan diff.
* Safely handles nil expiration via comma-ok type assertion. Previously empty strings assertions and null assertions were resulted in different hash values.

DiffSuppressFunc on both fields prevents plan-level diffs, and matches() uses case-insensitive comparison to avoid unnecessary update API calls.

### Check list
* [X] Describe the changes and their purpose
* [ ] Update HTML docs (if applicable)
    * [preview tool](https://registry.terraform.io/tools/doc-preview)
* [X] Update [`CHANGELOG.md`](../CHANGELOG.md)
* [ ] Check [`CONTRIBUTING.md`](../CONTRIBUTING.md) for anything forgotten

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## X.Y.Z (Unreleased)
+BUG FIXES:
+* Updated `sumologic_cse_match_list` to handle cases where case differences in values and empty expirations resulted in plan diffs with no functional changes. These will no longer result in a plan diff.
 
 ## 3.2.10 (Jul 15, 2026)
 FEATURES:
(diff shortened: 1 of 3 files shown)


rsha-sumo:
looks fine to me

rsha-sumo (approved):

PR #892, opened 2026-08-21
Add content argument to sumologic_lookup_table for CSV uploads

### LINKED ISSUE #202: Updating lookup CSV content via Sumologic provider in Terraform
Hi, I noticed that it is possible to create a new lookup definition but is there a way to specify the contents of the lookup table (e.g. via CSV) through the Sumologic Provider in Terraform? 

I looked at [sumologic_lookup_table](https://registry.terraform.io/providers/SumoLogic/sumologic/latest/docs/resources/lookup_table#attributes-reference) resource, but did not see any option to specify a local file/path which contains the contents of the lookup table to sync with Sumologic.

arminsumo:
## Summary
- Closes #202. Adds a `content` string argument to `sumologic_lookup_table` (populate via `file("data.csv")`), which uploads CSV rows through Sumo Logic's lookup table upload/truncate APIs on create/update.
- CSV structure (headers vs. `fields`/`primary_keys`, ragged rows, unbalanced quotes) is validated at `terraform plan` time via `CustomizeDiff`, before any API call.
- State stores only a sha256 hash of `content`, never the raw CSV, since the API has no way to read rows back.
- Size guard: errors above Sumo Logic's 100MB CSV upload limit, warns above 10MB.

## Details
- `sumologic/sumologic_client.go`: new `PostMultipartFile` helper (multipart/form-data POST), reusing the existing request/response pipeline.
- `sumologic/sumologic_lookup_table.go`: new `UploadLookupTableContent`, `TruncateLookupTable`, and a dedicated `waitForLookupJob` poller. The existing `waitForJob` helper can't be reused as-is — lookup jobs report `Pending`/`PartialSuccess` states it would treat as errors, and its `Status` struct doesn't carry the `errors[]`/`warnings[]` detail lookup jobs return.
- `sumologic/resource_sumologic_lookup_table.go`: `content` schema field (`StateFunc` hashes it, `ValidateFunc` checks size/line-endings), a `Timeouts` block (the resource had none), and `CustomizeDiff` wiring the plan-time CSV structure check. Removing `content` truncates the table on update rather than leaving stale rows.
- Docs (`website/docs/r/lookup_table.html.markdown`) and `CHANGELOG.md` updated.

Deliberately out of scope (documented in code comments): cell-value validation against `field_type` (would duplicate server-side coercion this provider can't observe), a `merge` toggle (Terraform's declarative model is full-replace), and `content_file`/`content_hash` args (additive if a real oversized-CSV case shows up; `file()` covers the realistic range under the API's 100MB ceiling).

## Test plan
- [x] `go vet ./sumologic/...` and `go install ./...` clean
- [x] `go test ./sumologic/ -run 'TestHashLookupContent|TestValidateLookupContent|TestValidateLookupCSV' -v` — 15/15 pass, including ragged-row and unbalanced-quote cases
- [x] `TF_ACC=1 go test ./sumologic -run 'TestAccSumologicLookupTable' -v` against a real org — all 4 pass (`_basic`, `_create`, `_update`, `_content`), each creates and destroys its own table
- [x] Manual e2e against the same real org, one throwaway table, config-driven (no repo changes):
  - create with `content` → API confirms non-zero stored size
  - re-plan with unchanged `content` → no diff (hash is stable, no needless re-upload)
  - swap `content` to a different CSV → exactly 1 changed, stored size reflects the new bytes
  - CSV header missing a declared field → rejected at `terraform plan`, before any API call
  - `cat path://"<table path>"` search → returned the exact row uploaded (`{"key":"alpha","value":"one"}`), confirming rows are genuinely queryable, not just accepted
  - remove `content` → stored size drops to 0 (truncate-on-removal)
  - `terraform destroy` → table gone (API 404/400)
  - org diffed before/after: personal folder contents identical, no leftover tables or other content

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 ## X.Y.Z (Unreleased)
+ENHANCEMENTS:
+* `sumologic_lookup_table`: Added `content` argument to populate a lookup table's rows with CSV data (e.g. via `file()`), addressing [#202](https://github.com/SumoLogic/terraform-provider-sumologic/issues/202).
 
 ## 3.3.1 (Aug 20, 2026)
 FEATURES:
--- a/sumologic/resource_sumologic_lookup_table.go
+++ b/sumologic/resource_sumologic_lookup_table.go
@@ -1,22 +1,40 @@
 package sumologic
 
 import (
+	"context"
+	"crypto/sha256"
+	"encoding/csv"
+	"encoding/hex"
 	"fmt"
+	"io"
 	"log"
+	"strings"
+	"time"
 
+	"github.com/hashicorp/terraform-plugin-sdk/v2/diag"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/schema"
 	"github.com/hashicorp/terraform-plugin-sdk/v2/helper/validation"
 )
 
+// Sumo Logic's documented limits for the lookup table CSV upload API
+// (https://api.sumologic.com/docs/#operation/uploadFile). maxContentBytes is
+// enforced as a hard error; warnContentBytes is a softer nudge that large,
+// slow-changing data belongs in an ingest pipeline rather than Terraform.
+const (
+	maxContentBytes  = 100 * 1024 * 1024
+	warnContentBytes = 10 * 1024 * 1024
+)
+
 func resourceSumologicLookupTable() *schema.Resource {
 	return &schema.Resource{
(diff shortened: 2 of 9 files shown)


ndhar-sumo (approved):
lgtm

arminsumo:
Superseded by #893 — moved in-repo so the acceptance suite runs with credentials (fork PRs get no repository secrets).

